PR #2, opened 2020-04-18
Fix proxy headers

Temporalin:
Proxy wasn't working for me because the headers had default values from the `requests` library, so I fixed it by using a `PreparedRequest`

### DIFF
--- a/bemani/utils/proxy.py
+++ b/bemani/utils/proxy.py
@@ -166,13 +166,14 @@ def receive_request(path: str) -> Response:
 
     # Make request to foreign service, using the same parameters
     prep_req = requests.Request(
-		'POST',
-		url=f'http://{remote_host}:{remote_port}{actual_path}',
-		headers=headers,
-		data=req_binary,
+        'POST',
+        url=f'http://{remote_host}:{remote_port}{actual_path}',
+        headers=headers,
+        data=req_binary,
     ).prepare()
-	
-	r = sess.send(prep_req)
+
+    sess = requests.Session()
+    r = sess.send(prep_req)
 
     if r.status_code != 200:
         # Failed on remote side
--- a/bemani/utils/proxy.py
+++ b/bemani/utils/proxy.py
@@ -154,7 +154,7 @@ def receive_request(path: str) -> Response:
         )
 
     # Set up custom headers for remote request
-    headers = {}
+    headers = {key: value for key, value in request.headers.items()}
     if request_compression is not None:
         headers['X-Compress'] = request_compression
     if request_encryption is not None:
@@ -165,12 +165,14 @@ def receive_request(path: str) -> Response:
     headers['X-Remote-Address'] = remote_address or request.remote_addr
 
     # Make request to foreign service, using the same parameters
-    r = requests.post(
-        f'http://{remote_host}:{remote_port}{actual_path}',
-        headers=headers,
-        data=req_binary,
-        timeout=config['timeout'],
-    )
+    prep_req = requests.Request(
+		'POST',
+		url=f'http://{remote_host}:{remote_port}{actual_path}',
+		headers=headers,
+		data=req_binary,
+    ).prepare()
+	
+	r = sess.send(prep_req)
 
     if r.status_code != 200:
         # Failed on remote side


DragonMinded:
I'm not sure what exactly wasn't working for you. This seems to break the timeout handling and passes on a bunch of headers that aren't needed for the proxy request to work in the first place. What was going wrong?

bemani/utils/proxy.py
```diff
@@ -154,7 +154,7 @@ def receive_request(path: str) -> Response:
         )
 
     # Set up custom headers for remote request
-    headers = {}
+    headers = {key: value for key, value in request.headers.items()}
```
DragonMinded, line 157: All of the headers that are needed for a successful XRPC request are set below in the if statements.

bemani/utils/proxy.py
```diff
@@ -165,12 +165,14 @@ def receive_request(path: str) -> Response:
...
-        timeout=config['timeout'],
-    )
+    prep_req = requests.Request(
+		'POST',
+		url=f'http://{remote_host}:{remote_port}{actual_path}',
+		headers=headers,
+		data=req_binary,
+    ).prepare()
+	
+	r = sess.send(prep_req)
```
DragonMinded, line 175: What does splitting this accomplish?

Temporalin:
> 
> 
> I'm not sure what exactly wasn't working for you. This seems to break the timeout handling and passes on a bunch of headers that aren't needed for the proxy request to work in the first place. What was going wrong?

The server I'm using returned a 500 error until I did this. Inspecting the requests I found that the `requests` library added a bunch of headers (I can check them again if needed, but for example there was one for compression with values like gzip that was not in the request made by the game) unless I did it step by step like I do.

The headers you mention are indeed correct, but I think the problem relies on the other ones.

I forgot to set the timeout but I'm sure that can be added to this fix.

DragonMinded:
I'm looking at a request right now, the only headers that are being sent by the game include:

 - Content-Length (for the request binary data)
 - Host (for SNI, but older games put the server's IP in here so it is not reliable)
 - X-Compress (whether to use LZ77 or no compression)
 - X-Eamuse-Info (whether to use RC4 or no encryption)
 - User-Agent (Just the XRPC client in the game, should be droppable)
 - Keep-Alive (Again, shouldn't be required)

My guess is that the server is erroring due to the incorrect omission of Content-Length, Host or both. Possibly Host due to being on a virtual server. I recommend checking for the existence of Content-Length, Host and User-Agent and passing those forward explicitly and seeing if that doesn't fix things.

DragonMinded:
Did you get a chance to try the suggested changes?

Temporalin:
I'll check again this weekend and post again what I find

Temporalin:
These are the headers received by the proxy
```
User-Agent: EAMUSE.Httpac/1.0
Content-Length: 164
X-Eamuse-Info: 1-0123abcd-4567
Host: <local IP address of machine running game>
X-Compress: none
```

These are the headers sent by the proxy without my modifications, which is responded with a 500 error
```
Host: <local IP address of machine running proxy>
User-Agent: python-requests/2.23.0
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
X-Compress: none
X-Eamuse-Info: 1-0123abcd-4567
X-Remote-Address: <local IP address of machine running game>
Content-Length: 164
```

And finally, these are the headers sent by the proxy after my modifications
```
Accept-Encoding: identity
User-Agent: EAMUSE.Httpac/1.0
Content-Length: 164
X-Eamuse-Info: 1-0123abcd-4567
Host: <local IP address of machine running proxy>
X-Compress: none
X-Remote-Address: <local IP address of machine running game>
```

These request is the first HTTP request the game sends. After running successfuly, the game sends some more requests and the game runs correctly.

Maybe the problem relies on a bad server implementation?

DragonMinded:
Now that I'm thinking about it, the "Host" field should not be modified (requests will fill the right one in based on the domain/IP configured in the proxy), and "Content-Length" should not be either (requests will fill in the right value based on its modified contents). "X-Remote-Address" allows servers running this code fronted by one or more proxies to still detect the connecting IP, which is a big deal for matching and why I added it. This leaves basically the "User-Agent" which should definitely be cloned from the requesting client, so that servers which care about the requesting game and switch on this (not this one, but possible some) get the right information.

"Connection", "Accept-Encoding" and "Accept" shouldn't be forwarded at all, since they aren't in the original spec. However, I suspect that the remote server might not respond correctly to a genuine request. What happens if you point the game directly at the server? Does it actually run, or does it 500? Do you know what the server is running on? The recommendation with this server is to front with nginx, which will handle all sorts of bizarre cases, but I've seen implementations that are 100% in Java/C++ and might not handle full range of HTTP.

My guess is that it doesn't like the "Accept-Encoding" and only supports basic identity requests which should be the default assumption if a server doesn't get a header for that. It seems a reasonable thing for a server to 500 if you state that you only accept encodings that it does not support. What happens if you modify nothing, except to set "Accept-Encoding" to "identity"?

DragonMinded:
It looks like the only way to suppress sending these headers is with an adapter, such as https://github.com/kanzure/careful-requests

DragonMinded:
I can't find full documentation, but it appears using a PreparedRequest sidesteps adding a bunch of unnecessary headers, so that's probably why this is working. It still has the issue of copying all request headers over before explicitly copying some of them again in the series of if statements, and the missing timeout. I proposed an alternate fix in #3 but I can't test, do you want to have a look at that, or fix those issues listed?

Temporalin:
The fix in #3 works just fine and is already cleaner code, so I would merge your PR and close this one.

DragonMinded:
Cool, thanks for the report and the help in testing!

PR #6, opened 2020-08-18
Force decoding JSX files as utf-8 regardless of the OS locale

DragonMinded:
Its possible that the OS settings means that reading jsx files as string means that there is a possible decode error. I am not sure this will fix all issues, but it should at least make reading and decoding the file before passing it to the jsx compiler the same on all systems.


DragonMinded:
This is already in trunk.

PR #8, opened 2020-10-17
Initial support of 2dx26

plxty:
Based on 2019090200.

## What's working?
1. read song database
2. free mode with card inserted

## What's not?
1. arena, other modes, etc.
2. still need further tests

### DIFF
--- a/bemani/backend/iidx/factory.py
+++ b/bemani/backend/iidx/factory.py
@@ -28,6 +28,7 @@ from bemani.backend.iidx.pendual import IIDXPendual
 from bemani.backend.iidx.copula import IIDXCopula
 from bemani.backend.iidx.sinobuz import IIDXSinobuz
 from bemani.backend.iidx.cannonballers import IIDXCannonBallers
+from bemani.backend.iidx.rootage import IIDXRootage
 from bemani.common import Model, VersionConstants
 from bemani.data import Data
 
@@ -60,6 +61,7 @@ class IIDXFactory(Factory):
         IIDXCopula,
         IIDXSinobuz,
         IIDXCannonBallers,
+        IIDXRootage,
     ]
 
     @classmethod
@@ -81,9 +83,9 @@ class IIDXFactory(Factory):
                 return VersionConstants.IIDX_COPULA
             if date >= 2016102600 and date < 2017122100:
                 return VersionConstants.IIDX_SINOBUZ
-            if date >= 2017122100:
+            if date >= 2017122100 and date < 2019012300:
                 return VersionConstants.IIDX_CANNON_BALLERS
-            return None
+            return VersionConstants.IIDX_ROOTAGE # date >= 2019012300
 
         if model.game == 'JDJ':
             return IIDXSirius(data, config, model)
@@ -128,6 +130,8 @@ class IIDXFactory(Factory):
                 return IIDXSinobuz(data, config, model)
             if version == VersionConstants.IIDX_CANNON_BALLERS:
                 return IIDXCannonBallers(data, config, model)
+            if version == VersionConstants.IIDX_ROOTAGE:
+                return IIDXRootage(data, config, model)
 
         # Unknown game version
         return None
--- a/bemani/backend/iidx/rootage.py
+++ b/bemani/backend/iidx/rootage.py
@@ -1,0 +1,1299 @@
+# vim: set fileencoding=utf-8
+import copy
+import struct
+from typing import Optional, Dict, Any
+
+from bemani.backend.iidx.base import IIDXBase
+from bemani.backend.iidx.sinobuz import IIDXSinobuz
+from bemani.backend.iidx.cannonballers import IIDXCannonBallers
+
+from bemani.common import ValidatedDict, Model, VersionConstants, Time, ID, intish
+from bemani.data import Data, UserID
+from bemani.protocol import Node
+
+# todo: clear '?' mark
+class IIDXRootage(IIDXBase):
+
+    name = 'Beatmania IIDX ROOTAGE'
+    version = VersionConstants.IIDX_ROOTAGE
+    sinobuz = None
+
+    # todo: concurrency?
+    def s(self) -> IIDXSinobuz:
+        if self.sinobuz is None:
+            self.sinobuz = IIDXSinobuz(self.data, self.config, self.model)
+        return self.sinobuz
+
+
+    # proxied: sinobuz.py
+    def db_to_game_status(self, db_status: int) -> int:
+        return self.s().db_to_game_status(db_status)
+    def game_to_db_status(self, game_status: int) -> int:
+        return self.s().game_to_db_status(game_status)
+    def db_to_game_rank(self, db_dan: int, cltype: int) -> int:
+        return self.s().db_to_game_rank(db_dan, cltype)
+    def game_to_db_rank(self, game_dan: int, cltype: int) -> int:
+        return self.s().game_to_db_rank(game_dan, cltype)
+
+
+    def previous_version(self) -> Optional[IIDXBase]:
+        return IIDXCannonBallers(self.data, self.config, self.model)
+
+
+    def format_profile(self, userid: UserID, profile: ValidatedDict) -> Node:
+        # modified from sinobuz.py
+        root = Node.void('IIDX26pc')
+
+        # Look up play stats we bridge to every mix
+        play_stats = self.get_play_statistics(userid)
+
+        # Look up judge window adjustments
+        judge_dict = profile.get_dict('machine_judge_adjust')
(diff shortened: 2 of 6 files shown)


bemani/backend/iidx/factory.py
```diff
@@ -81,9 +83,9 @@ def version_from_date(date: int) -> Optional[int]:
                 return VersionConstants.IIDX_COPULA
             if date >= 2016102600 and date < 2017122100:
                 return VersionConstants.IIDX_SINOBUZ
-            if date >= 2017122100:
+            if date >= 2017122100 and date < 2019012300:
                 return VersionConstants.IIDX_CANNON_BALLERS
-            return None
+            return VersionConstants.IIDX_ROOTAGE # date >= 2019012300
```
DragonMinded, line 88: Prefer putting this in an if statement and continuing to return None at the bottom as a safeguard against future modifications accidentally defaulting to rootage.

bemani/backend/iidx/rootage.py
```diff
@@ -0,0 +1,1299 @@
...
+from bemani.common import ValidatedDict, Model, VersionConstants, Time, ID, intish
+from bemani.data import Data, UserID
+from bemani.protocol import Node
+
+# todo: clear '?' mark
+class IIDXRootage(IIDXBase):
+
+    name = 'Beatmania IIDX ROOTAGE'
+    version = VersionConstants.IIDX_ROOTAGE
+    sinobuz = None
```
DragonMinded, line 19: Why do we need a handle specifically to sinobuz? If you need old version support, you should use the `previous_version` paradigm that's used across the board.

bemani/backend/iidx/rootage.py
```diff
@@ -0,0 +1,1299 @@
...
+
+    # todo: concurrency?
+    def s(self) -> IIDXSinobuz:
+        if self.sinobuz is None:
+            self.sinobuz = IIDXSinobuz(self.data, self.config, self.model)
+        return self.sinobuz
+
+
+    # proxied: sinobuz.py
```
DragonMinded, line 28: Prefer copypasta over proxy, this was intentionally chosen to bias towards ability to change individual games without unintentially breaking forwards or backwards compatibility with other games. Its fine to have identical code, since the opposite end is something we don't control and the code is better documenting as to what's going on in that particular game version.

bemani/backend/iidx/rootage.py
```diff
@@ -0,0 +1,1299 @@
...
+        judge_dict = profile.get_dict('machine_judge_adjust')
+        machine_judge = judge_dict.get_dict(self.config['machine']['pcbid'])
+
+        # Profile data
+        pcdata = Node.void('pcdata')
+        pcdata.set_attribute('id', str(profile.get_int('extid')))
+        pcdata.set_attribute('idstr', ID.format_extid(profile.get_int('extid')))
+        pcdata.set_attribute('name', profile.get_str('name'))
+        pcdata.set_attribute('pid', str(profile.get_int('pid')))
+        pcdata.set_attribute('spnum', str(play_stats.get_int('single_plays')))
```
DragonMinded, line 60: I see IIDX still hasn't moved away from attributes when they have an entire type system they're ignoring...

bemani/backend/iidx/rootage.py
```diff
@@ -0,0 +1,1299 @@
...
+        achievement_node.set_attribute('weekly_num', str(profile.get_int('weekly_num')))
+        # Prefecture visit flag
+        achievement_node.set_attribute('visit_flg', str(profile.get_int('visit_flg')))
+        # Number of rivals beaten
+        achievement_node.set_attribute('rival_crush', str(profile.get_int('rival_crush')))
+        # Tran medals
+        achievement_node.add_child(Node.s64_array('trophy', profile.get_int_array('trophy', 10)))
+        root.add_child(achievement_node)
+
+        # Track deller
```
DragonMinded, line 423: Is it still referred to as deller?

bemani/backend/iidx/rootage.py
```diff
@@ -0,0 +1,1299 @@
...
+
+            # fixme: monthly_mranking, total_mranking
+            # todo: internet ranking (i cannot find any infomation about this)
+            # todo: cm, what is it
+
+            # KONAMI Arcade Championship 8th
+            root.add_child(Node.string('kac_mid', '26033'))
+            root.add_child(Node.string('kac_clid', str(self.CHART_TYPE_A7))) # not sure :/
+
+            # others are from sinobuz.py, ignored omnimix
```
DragonMinded, line 755: You'll want to copy the way omni/non-omni flags work from previous mixes.

bemani/backend/iidx/rootage.py
```diff
@@ -0,0 +1,1299 @@
...
+            # expire
+            root.set_attribute('expire', '600')
+
+            # fixme: monthly_mranking, total_mranking
+            # todo: internet ranking (i cannot find any infomation about this)
+            # todo: cm, what is it
+
+            # KONAMI Arcade Championship 8th
+            root.add_child(Node.string('kac_mid', '26033'))
+            root.add_child(Node.string('kac_clid', str(self.CHART_TYPE_A7))) # not sure :/
```
DragonMinded, line 753: You can usually get this information from bemaniwiki 2nd.

bemani/backend/iidx/rootage.py
```diff
@@ -0,0 +1,1299 @@
...
+    def handle_IIDX26shop_request(self, request: Node) -> Node:
+        method = request.attribute('method')
+        root = Node.void('IIDX26shop')
+
+        # 0x18067D020
+        if method == 'sentinfo':
+            pass # ignore
+
+        # 0x18067D0B0
+        if method == 'getname':
```
DragonMinded, line 1134: What happened to 'savename'?

bemani/format/iidxmusicdb.py
```diff
@@ -80,6 +80,9 @@ def __parse_db(self, data: bytes) -> None:
         elif data[4] == 0x18:
             offset = 0xc360
             leap = 0x340
+        elif data[4] == 0x1a:
```
DragonMinded, line 83: This is such a gross hack on my part. This number is clearly usable for calculating offsets, but oh well.

requirements.txt
```diff
@@ -12,7 +12,7 @@ freezegun
 pyreact
 Flask-Caching
 blinker
-pycrypto
+pycryptodome
```
DragonMinded, line 15: What is the difference here?

DragonMinded (commented):
Thanks for taking a stab at this! I've peppered it with comments, sorry about that. In general it looks like a decent amount was copied from sinobuz (which is how I've done it as well), have you tested the various endpoints and seen what the game actually needs versus what's vestigial from old versions? Also, any reason for skipping cannon ballers and only supporting non-omni? It leaves some interesting holes in the software.

bemani/backend/iidx/factory.py
plxty: You're right, I havn't considered that much, thanks ;)

plxty:
Thanks for reviewing😀

1. Yes, most of the codes are came from `sinobuz.py`, I'm not familiar with the easrv honestly, so I left the part that needs some logic calculations, and only make it "works" for now😂
2. I'm playing 26 personally, and only tested in free mode (I'm a freshman to 2dx)
3. As above, I have no 25 and omni version; btw is that 26-omni means 25-omni plus the 26 songs?

This is just an initial version, it still has a lot of things to do. Thus I'm looking forward to any chances like this, thanks, appreciatively😉

bemani/backend/iidx/rootage.py
plxty: Yes it's an instance that proxied for avoding duplicate codes. I'm agreed with you that we need to reduce the coupling of two modules.

bemani/backend/iidx/rootage.py
plxty: Is that a typo? Interesting😂

bemani/backend/iidx/rootage.py
plxty: Still can't find the chart information🤔

bemani/backend/iidx/rootage.py
plxty: The callback function that decompiled doesn't contains the `savename` field, what was its functionality previously?

requirements.txt
plxty: `pycrypto` is quite an old thing, and it refuse to compile on my Windows machine. Therefore I've switched to that forked version.

Subject38:
Now I'm tempted to submit a massive pull request with all the stuff i've added but my code is an absolute mess and doesn't have any unit tests or anything so idk :/

DragonMinded:
Closing old pull requests that have no activity.

PR #9, opened 2020-11-07
Support IIDX 25-27, jubeat festo, pop'n music peace, and SDVX Vivid Wave

Subject38:
I nuked the repository and I'm going to reopen this as a set of multiple pull requests so that things can be better organized and so that we might actually have a chance of getting any of my changes mainlined

### DIFF
--- a/bemani/backend/core/eacoin.py
+++ b/bemani/backend/core/eacoin.py
@@ -131,7 +131,7 @@ class PASELIHandler(Base):
                 return make_resp(2, 0)
 
             userid = self.data.local.user.from_session(session)
-            if session is None:
+            if userid is None:
                 # Refuse to do anything
                 print("Invalid session for eacoin consume request")
                 return make_resp(2, 0)


Subject38 (commented):
I added the readme.md changes because I had this repo on a different site outside github

DragonMinded:
This is .....large. Its going to take me awhile to get through reviewing things. You said you didn't have any unit tests. Does that mean you don't have any game captures replayed in the client traffic tests? That's gotta be there before I can merge anything since its basically the last defense against regressions.

This is really cool, by the way! How seriously do you want to get it merged? We can definitely work on everything and get it mainlined if you want.

README.md
```diff
@@ -1,3 +1,5 @@
+This is a fork of DragonMinded's bemaniutils: https://github.com/DragonMinded/bemaniutils
```
DragonMinded: Not anymore ;)

bemani/api/objects/catalog.py
```diff
@@ -165,11 +166,8 @@ def __format_extras(self) -> Dict[str, List[Dict[str, Any]]]:
 
     @property
     def music_version(self) -> int:
-        if self.game == GameConstants.IIDX:
-            if self.omnimix:
-                return self.version + DBConstants.OMNIMIX_VERSION_BUMP
-            else:
-                return self.version
+        if self.omnimix:
```
DragonMinded: The implication of this change is that it standardizes omnimixes across all games, instead of just for IIDX. I think I'm fine with that. I kept this IIDX-specific because there's no omnimix support for other mixes, so if a remote server requests songs for an omnimix version of another game, they would get nothing. Are we okay with that implication (mapping requests to omnimixes for non-IIDX to empty responses, instead of the base game's responses)?

bemani/api/objects/catalog.py
```diff
@@ -183,7 +181,7 @@ def fetch_v1(self, idtype: str, ids: List[str], params: Dict[str, Any]) -> Dict[
 
         # Fetch the songs
         songs = self.data.local.music.get_all_songs(self.game, self.music_version)
-        if self.game == GameConstants.JUBEAT and self.version == VersionConstants.JUBEAT_CLAN:
+        if self.game == GameConstants.JUBEAT and self.version == VersionConstants.JUBEAT_CLAN or self.version == VersionConstants.JUBEAT_FESTO:
```
DragonMinded: Needs parens for order of operations, not sure what the intent is here. Does Festo also have the same song range?

bemani/api/objects/records.py
```diff
@@ -89,6 +89,7 @@ def __format_jubeat_record(self, record: Score) -> Dict[str, Any]:
             'status': status,
             'combo': record.data.get_int('combo', -1),
             'ghost': ghost,
+            'music_rate': record.data.get_int('music_rate', -1)
```
DragonMinded: This will need a corresponding update to the API specification. What's music_rate used for?

bemani/backend/iidx/base.py
```diff
@@ -38,9 +38,25 @@ class IIDXBase(CoreHandler, CardManagerHandler, PASELIHandler, Base):
     CHART_TYPE_N14 = 3
     CHART_TYPE_H14 = 4
     CHART_TYPE_A14 = 5
-    # Beginner charts only save status
+    # Beginner charts only save status(not anymore lol)
```
DragonMinded: Should be updated to reflect that Beginner charts only save status up though <release>

Subject38:
I don't really know much about  how you designed the testing framework so I'm gonna say no to that first question. As for actually getting it merged, I would be interested, I have no idea what kind of time span we're talking because I have university and stuff but I would love to get the codebase to a point that it actually fits with everything else. I've mostly been working on this for a production instance that I've been hosting for some friends so that's why I've almost exclusively focused purely on backend stuff/making the games work.

bemani/backend/iidx/base.py
```diff
@@ -38,9 +38,25 @@ class IIDXBase(CoreHandler, CardManagerHandler, PASELIHandler, Base):
...
     CHART_TYPE_B7 = 6
-
+    # HV and above include all the difficulties in one songid now
+    CHART_TYPE_L7 = 7
+    CHART_TYPE_B14 = 8
+    CHART_TYPE_L14 = 9
+    # Used for dealing with mismatch of chart ids from HV and onwards
```
DragonMinded: What is this doing here? The whole point of having indirection between DB chart and individual game chart mappings is that this is allowed to update, change, get juggled, etc. There shouldn't be a need for mapping updates in the base. See DDR for an example of how songs got completely renumbered, and old SDVX for when they renumbered chart types for some examples.

bemani/backend/iidx/cannonballers.py
```diff
@@ -1,15 +1,2018 @@
...
+    def handle_IIDX25music_crate_request(self, request: Node) -> Node:
+        root = Node.void('IIDX25music')
+        attempts = self.get_clear_rates()
+
+        all_songs = list(set([song.id for song in self.data.local.music.get_all_songs(self.game, self.music_version)]))
+        for song in all_songs:
+            clears = []
+            fcs = []
+
+            for chart in [0, 1, 2, 3, 4, 5]:
```
DragonMinded: I think the mistake was that IIDX originally did not have any indirection in charts (jubeat might not either). We will need to introduce that in the existing IIDX versions to be correct.

bemani/backend/iidx/cannonballers.py
```diff
@@ -1,15 +1,2018 @@
...
+                'songs': [
+                    6013,
+                    22016,
+                    20073,
+                    24095,
+                ],
+            },
+        ]
+
+        # For some reason, omnimix crashes on course mode, so don't enable it
```
DragonMinded: Is this *still* true for CB+ omni? LOL

bemani/backend/iidx/heroicverse.py
```diff
@@ -0,0 +1,1883 @@
...
+            if cltype == self.CLEAR_TYPE_SINGLE:
+                if chart not in [
+                    self.CHART_TYPE_B7,
+                    self.CHART_TYPE_N7,
+                    self.CHART_TYPE_H7,
+                    self.CHART_TYPE_A7,
+                    self.CHART_TYPE_L7,
+                ]:
+                    continue
+                chartindex = {
```
DragonMinded: My mistake originally on all of this software (IIDX was one of the earliest ones I did and so you can see some old paradigms in the code), but this should be in a lookup function similar to the game_to_db_rank/db_to_game_rank functions above. The other mapping in base should move here, and we need to update the games before HV to include their own copypasta constants with unit mappings.

bemani/backend/iidx/heroicverse.py
```diff
@@ -0,0 +1,1883 @@
...
+                    self.CHART_TYPE_B14: 0,
+                    self.CHART_TYPE_N14: 1,
+                    self.CHART_TYPE_H14: 2,
+                    self.CHART_TYPE_A14: 3,
+                    self.CHART_TYPE_L14: 4,
+                }[chart]
+
+            if musicid not in scorestruct:
+                scorestruct[musicid] = [
```
DragonMinded: Good to see IIDX never used their typed xml system, not never.

bemani/backend/iidx/heroicverse.py
```diff
@@ -0,0 +1,1883 @@
...
+        expert.set_attribute('phase', '1')
+
+        expert_random_select = Node.void('expert_random_secret')
+        root.add_child(expert_random_select)
+        expert_random_select.set_attribute('phase', '1')
+
+        expert_full = Node.void('expert_secret_full_open')
+        root.add_child(expert_full)
+
+        # some new nodes for rootage
```
DragonMinded: We should hook these up to options like many other games have.

bemani/backend/iidx/heroicverse.py
```diff
@@ -0,0 +1,1883 @@
...
+            qprodata.set_attribute('head', str(qpro.get_int('head')))
+            qprodata.set_attribute('hair', str(qpro.get_int('hair')))
+            qprodata.set_attribute('face', str(qpro.get_int('face')))
+            qprodata.set_attribute('body', str(qpro.get_int('body')))
+            qprodata.set_attribute('hand', str(qpro.get_int('hand')))
+
+            challenge = Node.void('challenge')
+            rival.add_child(challenge)
+
+        # Expert  (removed in rootage)
```
DragonMinded: Things removed outright can just be deleted. I leave commented out code where somebody coming by to implement a feature I missed will have a starting point.

bemani/backend/iidx/rootage.py
```diff
@@ -0,0 +1,1812 @@
...
+        expert.set_attribute('phase', '1')
+
+        expert_random_select = Node.void('expert_random_secret')
+        root.add_child(expert_random_select)
+        expert_random_select.set_attribute('phase', '1')
+
+        expert_full = Node.void('expert_secret_full_open')
+        root.add_child(expert_full)
+
+        # some new nodes for rootage
```
DragonMinded: Same comment here.

bemani/backend/jubeat/base.py
```diff
@@ -137,7 +153,7 @@ def get_scores_by_extid(self, extid: Optional[int]) -> Optional[Node]:
             return None
 
         userid = self.data.remote.user.from_extid(self.game, self.version, extid)
-        scores = self.data.remote.music.get_scores(self.game, self.version, userid)
+        scores = self.data.remote.music.get_scores(self.game, self.music_version, userid)
```
DragonMinded: Oooh are we getting omnimix support for jubeat?

bemani/backend/jubeat/base.py
```diff
@@ -172,10 +190,11 @@ def update_score(
             self.PLAY_MEDAL_EXCELLENT,
         ]:
             raise Exception(f"Invalid medal value {medal}")
-
+        if hard_mode:
+            chart += 3
```
DragonMinded: Prefer an explicit lookup to adding magic numbers. Makes the code searchable and doesn't bake an impossible-to-find assumption into the current numbering of charts. Also, the base game score saving feature should only be responsible for directly saving/loading a score based on the chart provided. The normal/hard calculation should move to the game and be kept in a game-specific mapping there.

bemani/backend/jubeat/base.py
```diff
@@ -223,13 +242,21 @@ def update_score(
             # Update the ghost regardless, but don't bother with it in history
             scoredata.replace_int_array('ghost', len(ghost), ghost)
 
+        if music_rate != 0:
```
DragonMinded: Prefer None instead of 0 for "value is not present".

bemani/backend/jubeat/clan.py
```diff
@@ -1105,6 +1105,7 @@ def format_scores(self, userid: UserID, profile: ValidatedDict, scores: List[Sco
 
         music = ValidatedDict()
         for score in scores:
+            chart = score.chart if score.chart < 3 else score.chart - 3
```
DragonMinded: Again, should be a specific game_to_db_chart/db_to_game_chart lookup, even if its a lot more code. That makes it searchable and prevents coupling of game features to math that works "for now".

bemani/backend/jubeat/clan.py
```diff
@@ -1113,21 +1114,26 @@ def format_scores(self, userid: UserID, profile: ValidatedDict, scores: List[Sco
             ex_cnt = data.get_int_array('ex_cnt', 3)
             points = data.get_int_array('points', 3)
 
+            # This means that we already assigned a value and it was greater than current
```
DragonMinded: Not sure I agree with the philosophy on this. Did clan separate the scores or was "hard mode" just a tracking flag? I know festo has them separate. Trying to figure out what to do about existing support. I think the change for adding hard mode is going to be where we spend a lot of focus, to ensure its right.

bemani/backend/jubeat/clan.py
```diff
@@ -1809,7 +1816,7 @@ def unformat_profile(self, userid: UserID, request: Node, oldprofile: ValidatedD
                     if flags & bit > 0:
                         medal = max(medal, mapping[bit])
 
-                self.update_score(userid, timestamp, songid, chart, points, medal, combo, ghost, stats)
+                self.update_score(userid, timestamp, songid, chart, points, medal, combo, ghost, stats, hard_mode=is_hard_mode)
```
DragonMinded: This is where base DB charts should be calculated based on the game, instead of passing hard_mode down. That's an implementation detail that the base game shouldn't care about.

bemani/backend/jubeat/common.py
```diff
@@ -65,14 +65,56 @@ def handle_lobby_check_request(self, request: Node) -> Node:
...
+        entrant_nr = Node.u32('entrant_nr', 1)
         entrant_nr.set_attribute('time', '0')
         data.add_child(entrant_nr)
+        waitlist = Node.void('waitlist')
+        waitlist.set_attribute('count', '0')
+        data.add_child(waitlist)
 
         return root
 
+    def handle_lobby_entry_request(self, request: Node) -> Node:
```
DragonMinded: This looks like you're implementing lobby support (YAY) for jubeat, but I don't see any backend lobby storage. Are we faking it out or is it correct and the state is all client-side?

bemani/backend/jubeat/festo.py
```diff
@@ -1,16 +1,1830 @@
...
+            'enabled': False,
+        },
+        101: {
+            'enabled': False,
+        }
+    }
+    JBOX_EMBLEM_NORMAL = 1
+    JBOX_EMBLEM_PREMIUM = 2
+
+    FIVE_PLAYS_UNLOCK_EVENT_SONG_IDS = set(range(80000301, 80000348))
```
DragonMinded: Is this still in festo?

bemani/backend/jubeat/festo.py
```diff
@@ -1,16 +1,1830 @@
...
+                    'whim': daily_songs[1],
+                },
+            ))
+
+            # Mark that we did some actual work here.
+            data.local.network.mark_scheduled(cls.game, cls.version, 'fc_challenge', 'daily')
+        return events
+
+    def __get_course_list(self) -> List[Dict[str, Any]]:
+        # Full course list found at http://bemaniwiki.com/index.php?jubeat%20festo/TUNE%20RUN
```
DragonMinded: Nice! I *HATE* transcribing courses.

bemani/backend/jubeat/prop.py
```diff
@@ -1342,6 +1343,7 @@ def format_scores(self, userid: UserID, profile: ValidatedDict, scores: List[Sco
 
         music = ValidatedDict()
         for score in scores:
+            chart = score.chart if score.chart < 3 else score.chart - 3
```
DragonMinded: Same comment here, as well as regarding hard mode scores.

bemani/backend/museca/museca1plus.py
```diff
@@ -152,15 +152,26 @@ def enable_event(eid: int) -> None:
...
-
-        enable_event(1)  # Extended pedal options
-        enable_event(83)  # Light start
-        enable_event(130)  # Curator rank
-        enable_event(195)  # Fictional curator
-        # Event 194 is continuation mode, but it doesn't seem to work on latest data.
-
-        enable_event(98)  # Mission mode
-        for evtid in [145, 146, 147, 148, 149]:
+        event_ids = [
```
DragonMinded: Oh wow look at all this!

bemani/backend/popn/peace.py
```diff
@@ -1,15 +1,1323 @@
...
+                get_time = achievement.data.get_int('get_time')
+
+                item = Node.void('item')
+                root.add_child(item)
+                # Item type can be 0-6 inclusive and is the type of the unlock/item.
+                # Item 0 is music unlocks. In this case, the id is the song ID according
+                # to the game. Unclear what the param is supposed to be, but i've seen
+                # seen 8 and 0. Might be what chart is available?
+                #
+                # Item limits are as follows:
```
DragonMinded: Is this all still true? Should it be updated?

bemani/common/card.py
```diff
@@ -200,9 +200,9 @@ class CardCipher:
 
     @staticmethod
     def __type_from_cardid(cardid: str) -> int:
-        if cardid[:2].upper() == 'E0':
```
DragonMinded: What caused you to change this? I'm not sure at all which is correct, but I hesitate to make changes to long-baked code that's known working.

bemani/utils/read.py
```diff
@@ -883,15 +884,15 @@ def available_charts(mask: int) -> Tuple[bool, bool, bool, bool, bool, bool]:
                     mask & 0x4000000 > 0,  # Battle hyper chart bit
                 )
         elif self.version == VersionConstants.POPN_MUSIC_USANEKO:
-            # Based on M39:J:A:A:2018101500
```
DragonMinded: Why earlier?

bemani/utils/read.py
```diff
@@ -979,6 +980,201 @@ def available_charts(mask: int) -> Tuple[bool, bool, bool, bool, bool, bool]:
                     True,  # Always a battle normal chart
                     mask & 0x4000000 > 0,  # Battle hyper chart bit
                 )
+        # elif self.version == VersionConstants.POPN_MUSIC_PEACE:
```
DragonMinded: Why commented out?

bemani/utils/read.py
```diff
@@ -1502,11 +1750,16 @@ def __gather_sound_files(self, directory: str) -> Dict[int, str]:
             for filename in filenames:
                 songid, extension = os.path.splitext(filename)
                 if extension == '.1' or extension == '.ifs':
-                    try:
-                        files[int(songid)] = os.path.join(directory, os.path.join(dirpath, filename))
-                    except ValueError:
-                        # Invalid file
-                        pass
+                    if filename != '12030-p0.ifs':
```
DragonMinded: Haha what bullshit is this working around now? Maybe a comment about what's going on...

bemani/utils/read.py
```diff
@@ -1592,6 +1869,77 @@ def __revivals(self, songid: int, chart: int) -> Optional[int]:
             if old_id is not None:
                 return old_id
 
+        legg_map = {
```
DragonMinded: I was worried about this for the merge ;__;

data/jubeat.tsv
```diff
@@ -142,7 +143,6 @@
 30000010	銀河	フジファブリック
 30000011	元気でやってるのかい?	イルリメ
 30000012	Kick It Out	ブンブンサテライツ
-30000012	Kick It Out	ブンブンサテライツ
```
DragonMinded: Nice catch.

data/jubeat.tsv
```diff
@@ -1108,12 +1120,171 @@
...
 80001015	アガット [ 2 ]	Bazole
-90000002	インドがパッカーン！！煩悩マハラジャドリーム	ARM × かたほとり ft. 柊木りお
-90000003	ロミとロボの宇宙飛行	onoken
+90000001	Samba Ramba	Y&Co. feat. Kanae Asaba
+90000002	インドがパッカーン！！煩悩マハラジャドリーム	ARM × かたほとり ft. 柊木りお
+90000003	ロミとロボの宇宙旅行	onoken
 90000004	Sahara	Hommarju
 90000005	Phantasmagoria	Maozon
 90000006	EMOTiON TRiPPER	sky_delta
+90000007	恋 No リセットゲーム!!!!!	"DJ Genki feat. ななひら,yukacco,藍月なくる"
```
DragonMinded: Woo! I *hate* transcribing this stuff too.

bemani/backend/sdvx/vividwave.py
```diff
@@ -0,0 +1,4135 @@
...
+        # This should be created on the fly for a lobby that we're in.
+        game.add_child(Node.u32('entry_id', 1))
+        return game
+
+    def handle_game_sv5_entry_e_request(self, request: Node) -> Node:
+        # Lobby destroy method, eid node (u32) should be used
+        # to destroy any open lobbies.
+        return Node.void('game')
+
+    def __get_skill_analyzer_seasons(self) -> Dict[int, str]:
```
DragonMinded: Are these still the right courses?

bemani/backend/sdvx/vividwave.py
```diff
@@ -0,0 +1,4135 @@
...
+        enable_event("CONTINUATION")
+        enable_event("TENKAICHI_MODE")
+        enable_event("SERIALCODE_JAPAN")
+        enable_event("DEMOGAME_PLAY")
+        enable_event("TOTAL_MEMORIAL_ENABLE")
+        enable_event("EVENT_IDS_SERIALCODE_TOHO_02")
+        enable_event("KONAMI_50TH_LOGO")
+        enable_event("PREMIUM_TIME_ENABLE")
+        # enable_event("SKILL_ANALYZER_ABLE")
+        # Uncomment when skill analyser courses are imported
```
DragonMinded: I'm guessing no ;P

DragonMinded (commented):
Okay, I did a *very quick* initial pass with some comments. I haven't had a chance to try any of this yet, but just wanted to start the discussion.

DragonMinded:
That's exactly what I was focused on for the first year or two of this as well. But you realize that you quickly get overwhelmed testing every game with every new major change to make sure you don't regress old versions. Hence, my hard requirement for myself (and pull requests) that new games are fully integration-tested before we claim working. With the breadth of support there's just no other way.

I'm beginning to understand a bit more about what you did with hard mode, and I am coming around to maybe its the right thing to do? Still gotta mull it over and make sure that we're not shooting ourselves in the foot data-wise. You can fix code but you can't undo a score overwrite in a production DB.

In general this looks like a really good foundation for all of these games. LOTS of my implementations were copy-pasta'd before verifying with actual games as well. Some comments need to be updated, and probably some assumptions need to be verified, but that's a perfectly fine way to go about things.

I don't have any real timeline for a merge either, so if it takes a long time, so be it. It would be better to have a reference implementation that more networks use over a few splintered forks even if it takes a few months to get there.

bemani/api/objects/catalog.py
Subject38: This was one of the changes i made when adding omnimix support for jubeat. I didn't fully understand why it hard assumed iidx

bemani/api/objects/catalog.py
Subject38: The answer to this would be yes. The exact same issue clan has exists in festo as well

bemani/api/objects/records.py
Subject38: music_rate is a percentage based on the number of goods, greats, and perfects you get in a song. It was added in festo and is required for its jubility calculation. It's stored as an int internally though and goes up to 120% because of hard mode

bemani/backend/iidx/base.py
Subject38: I figured placing it here would make sense since since this is how the songs are going to be from heroic verse onwards. Probably better to put them in each game individually though

bemani/backend/jubeat/base.py
Subject38: This was just the "cleanest" solution I could come up with. My philosophy was to put it here since it lets me implement hard mode saving in each game without worrying about the math there

bemani/backend/jubeat/clan.py
Subject38: The  game sends the hard mode flag as far back as saucer. From what I could tell, the "correct" behavior is to send the highest of the two scores rather than filter them outright

bemani/backend/jubeat/common.py
Subject38: I'm faking it out right now so that I'm not sitting in the online matching screen for 30 seconds. I'm planning on implementing it properly though

bemani/backend/museca/museca1plus.py
Subject38: MUSECA Plus modpack stuff. I asked Cammy what I need to do and he showed me this

bemani/common/card.py
Subject38: There were valid Felica ids that broke the server and I couldn't actually figure out what all the possible valid ones were so i just let it accept whatever.

data/jubeat.tsv
Subject38: Thank you for this list though. It was SUPER helpful for a different project i was working on that involved converting all the charts in the game to a format that I could play in a simulator

bemani/backend/iidx/cannonballers.py
Subject38: Good question. Lots of comments to clean up overall probably

bemani/utils/read.py
Subject38: This is the latest date code that I could personally find from any source I have. You probably have better sources than me lol

bemani/utils/read.py
Subject38: A friend of mine told me specifically to not extract this one because for some reason the leggendaria charts for POODLE were already stored in the actual chart ifs file. And this overall essentially checks if there's an add on ifs file and extracts that one instead. I'll add a comment though yeah

bemani/backend/sdvx/vividwave.py
Subject38: In my defense, transcribing those festo courses made me not want to do it again lmao

bemani/api/objects/catalog.py
DragonMinded: Basically it was to provide limited score support of the base game's scores when a game doesn't have omnimix (which was any game outside of IIDX). Its up for debate what we should do here, the spec is intentionally vague about behavior around this and I chose to try to do the best effort response over a technically correct but less useful one.

bemani/api/objects/catalog.py
DragonMinded: Okay, we should parenthesize the jubeat version portion then, or specify `self.version in (CLAN/FESTO)` for correctness.

```
>>> False and False or True
True
>>> False and (False or True)
False
```

bemani/api/objects/records.py
DragonMinded: Ah, good to know. The spec will need updating as well then (https://github.com/DragonMinded/bemapi)

bemani/backend/sdvx/vividwave.py
DragonMinded: Yeah, I never want to do it again either, but I did it for all the games that had courses ;_;

bemani/backend/iidx/base.py
DragonMinded: To clarify, the DB constants should go here in base, and the mapping between constants and their game counterparts (or ignoring charts we don't support for a game) should go in the individual game implementations. It's my fault there's no "GAME_CHART_TYPE_XXXX" for IIDX, they should have been separated like other games. But we have an opportunity to correct it now that you caught me in the act ;)

bemani/backend/jubeat/base.py
DragonMinded: Given hard mode's plumbing through all games, it might actually make sense to move some of it to the base game like you've done here. However, still no magic addition. I'd greatly prefer a mapping, since that's greppable in the code.

bemani/backend/jubeat/clan.py
DragonMinded: What about highest of other things? The score save method for every game does a merge of the best properties (combo indicator, ghost, score, rate, etc. So if you get a full combo but worse score, your record still shows your best score AND full combo) like the real network does, so your stored score is always the best of each individual thing. If we are going to return the best of either normal or hard mode scores, I think we need to do this merging here as well to preserve the functionality.

bemani/backend/jubeat/common.py
DragonMinded: Ah, I disabled it globally for the older games, and the newer ones I just hacked it out of the state machine, so I never see the wait.

bemani/backend/museca/museca1plus.py
DragonMinded: Is this safe to use on a vanilla Museca? Do the feature flags do anything there or is it pointless? I ask because this change itself seems a bit unrelated to the rest and could be broken out to its own pull request and merged faster. I have another museca player in the house so it would be fun to get some new things :3

bemani/common/card.py
DragonMinded: Hmm. I think that this isn't harmful. The only real thing could be bootleg cards that were not "E004" but started with "E0" and used the v1 cypher for the back printing? Those would render on the UI as the v2 cypher now, wouldn't they? Do you know what prefix the felica ids were? Shouldn't we whitelist those/add them to the v2 cypher instead of tightening the v1?

bemani/backend/jubeat/clan.py
Subject38: To be completely honest with you, I forgot about that edge case and now I realise I have no idea how to deal with that.

DragonMinded: Yeah, providing support for score portability backwards and forwards for games that never intended that sort of behavior all while preserving the expected behavior for each game is **HARD**. I've spent many sleepless nights getting that stuff just right for games. Believe me, its a royal pain whenever they throw a curveball like this. If you look at the data providers for fetching scores from remote servers as well as local mysql (when you have specified a remote bemapi-compatible server as a data source), there's code in there that does this merging so that if, for instance, you get a full-combo on your friend's network but have a higher score on yours, since they're linked it will still show correctly on your game. We might want to generalize that merging code into a library, call it from the API fetch methods, and then in jubeat older than festo fetch the normal and hard mode scores and do that merge before presenting to the client.

Generalizing those merges would be beneficial for another reason: people sometimes play on two cards but forget to add the second card to their profile before doing so, and then want to merge their two accounts. Merging two accounts is really the act of pointing all cards in account A to account B, dropping account A from the user tables, copying over all the history entries and then merging all of the high scores. Everything there is trivial, except merging. Hence, why there's no admin function for doing that merge (yet). If we had that generalized, it would be one step closer to implementing a true "merge two accounts".

Subject38:
image
while working on something else that I never did fix, I ran into this because of the verify lint script. Is there any particular reason why these are f strings?

DragonMinded:
> image
> while working on something else that I never did fix, I ran into this because of the verify lint script. Is there any particular reason why these are f strings?

Because my flake8 was out of date. Fixing now.

bemani/backend/jubeat/base.py
Subject38: I'm not super skilled with python, is None a valid value for an integer? If so I'll just change this and the default value for this to None

bemani/backend/jubeat/base.py
Subject38:
```python
if hard_mode:
    hard_mode_map = {
        self.CHART_TYPE_BASIC: self.CHART_TYPE_HARD_BASIC,
        self.CHART_TYPE_ADVANCED: self.CHART_TYPE_HARD_ADVANCED,
        self.CHART_TYPE_EXTREME: self.CHART_TYPE_HARD_EXTREME,
    }
    chart = hard_mode_map.get(chart)
```
Do you mean doing something like this to make it more transparent what is going on?

Subject38 (commented):
Not sure if there's a more elegant way to do this. This is a minor performance improvement from just sending both structures no matter what but when carding in with 3 rivals, it can still take as much as 30 seconds for card in to complete on my 2 core vps :/

Subject38:
I added reflesia because a friend of mine wanted to play reflec so I figured I would start working on that. reflesia is a weird case though because it pretty much kills every chart from previous games and then replaced them?? And it used the same internal ids so all the scores would transfer "correctly" but the charts would be different so I didn't know whether I should figure out which charts actually carried over or if I should just not worry about it... So far the only thing I got working on reflesia for now though is boot up and musicdb importing. I can't for the life of me figure out how to reverse engineer that god forsaken binary with its vftables so I'm at a bit of a standstill there lol

PR #10, opened 2020-12-06
Museca plus

Subject38:
Just a separate pull request with only the necessary changes to support MUSECA plus

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ build/
 .hgignore
 .venv/
 bemani.egg-info/
+server/
+emblems
+.vscode/
--- a/bemani/api/objects/catalog.py
+++ b/bemani/api/objects/catalog.py
@@ -165,11 +165,8 @@ class CatalogObject(BaseObject):
 
     @property
     def music_version(self) -> int:
-        if self.game == GameConstants.IIDX:
-            if self.omnimix:
-                return self.version + DBConstants.OMNIMIX_VERSION_BUMP
-            else:
-                return self.version
+        if self.omnimix:
+            return self.version + DBConstants.OMNIMIX_VERSION_BUMP
         else:
             return self.version
 
--- a/bemani/api/objects/catalog.py
+++ b/bemani/api/objects/catalog.py
@@ -165,8 +165,11 @@ class CatalogObject(BaseObject):
 
     @property
     def music_version(self) -> int:
-        if self.omnimix:
-            return self.version + DBConstants.OMNIMIX_VERSION_BUMP
+        if self.game in [GameConstants.IIDX, GameConstants.MUSECA]:
+            if self.omnimix:
+                return self.version + DBConstants.OMNIMIX_VERSION_BUMP
+            else:
+                return self.version
         else:
             return self.version
 
--- a/bemani/api/objects/records.py
+++ b/bemani/api/objects/records.py
@@ -222,13 +222,10 @@ class RecordsObject(BaseObject):
 
     @property
     def music_version(self) -> int:
-        if self.game == GameConstants.IIDX:
             if self.omnimix:
                 return self.version + DBConstants.OMNIMIX_VERSION_BUMP
             else:
                 return self.version
-        else:
-            return self.version
 
     def fetch_v1(self, idtype: str, ids: List[str], params: Dict[str, Any]) -> List[Dict[str, Any]]:
         since = params.get('since')
--- a/bemani/api/objects/records.py
+++ b/bemani/api/objects/records.py
@@ -222,8 +222,11 @@ class RecordsObject(BaseObject):
 
     @property
     def music_version(self) -> int:
-        if self.omnimix:
-            return self.version + DBConstants.OMNIMIX_VERSION_BUMP
+        if self.game in [GameConstants.IIDX, GameConstants.MUSECA]:
+            if self.omnimix:
+                return self.version + DBConstants.OMNIMIX_VERSION_BUMP
+            else:
+                return self.version
         else:
             return self.version
 
--- a/bemani/api/objects/records.py
+++ b/bemani/api/objects/records.py
@@ -222,10 +222,10 @@ class RecordsObject(BaseObject):
 
     @property
     def music_version(self) -> int:
-            if self.omnimix:
-                return self.version + DBConstants.OMNIMIX_VERSION_BUMP
-            else:
-                return self.version
+        if self.omnimix:
+            return self.version + DBConstants.OMNIMIX_VERSION_BUMP
+        else:
+            return self.version
 
     def fetch_v1(self, idtype: str, ids: List[str], params: Dict[str, Any]) -> List[Dict[str, Any]]:
         since = params.get('since')
--- a/bemani/api/objects/statistics.py
+++ b/bemani/api/objects/statistics.py
@@ -25,8 +25,11 @@ class StatisticsObject(BaseObject):
 
     @property
     def music_version(self) -> int:
-        if self.omnimix:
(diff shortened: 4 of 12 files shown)


Subject38:
image
idk how to fix this... (the blank box contains all the museca plus stuff but i want it all in the same folder as 1+1/2)

Subject38:
I ended fixing the display issue by hard coding the version name in the webui. Considering the fact that there won't be any more museca games and any other solution would likely involve making a new class just for museca plus, i figured this is a decent way to work around it

.gitignore
```diff
@@ -12,3 +12,6 @@ build/
 .hgignore
 .venv/
 bemani.egg-info/
+server/
```
DragonMinded, line 15: What is this?

.gitignore
```diff
@@ -12,3 +12,6 @@ build/
...
+emblems
```
DragonMinded, line 16: Soon >_>

bemani/api/objects/catalog.py
```diff
@@ -165,11 +165,8 @@ def __format_extras(self) -> Dict[str, List[Dict[str, Any]]]:
 
     @property
     def music_version(self) -> int:
-        if self.game == GameConstants.IIDX:
```
DragonMinded, line 168: I think we should change to `if self.game in {GameConstants.IIDX, GameConstants.MUSECA]` because otherwise we end up requesting and sending empty lists for other games that have no omni support if omnimix is requested. The current setup sends the base game's scores/etc which, while incomplete, is still better than nothing.

bemani/api/objects/records.py
```diff
@@ -222,11 +222,8 @@ def __format_record(self, cardids: List[str], record: Score) -> Dict[str, Any]:
 
     @property
     def music_version(self) -> int:
-        if self.game == GameConstants.IIDX:
```
DragonMinded, line 225: Same comment.

bemani/api/objects/statistics.py
```diff
@@ -25,11 +25,8 @@ def __format_user_statistics(self, cardids: List[str], stats: Dict[str, Any]) ->
 
     @property
     def music_version(self) -> int:
-        if self.game == GameConstants.IIDX:
```
DragonMinded, line 28: Same comment.

bemani/frontend/museca/endpoints.py
```diff
@@ -101,7 +101,7 @@ def viewnetworkrecords() -> Response:
     frontend = MusecaFrontend(g.data, g.config, g.cache)
     network_records = frontend.get_network_records()
     versions = {version: name for (game, version, name) in frontend.all_games()}
-    versions[0] = 'CS and Licenses'
```
DragonMinded, line 104: LOL copypasta from Pop'n. I wonder if there was a reason for it being here instead of in the `all_games` method. I should have wrote a comment...

bemani/frontend/museca/endpoints.py
```diff
@@ -139,6 +139,7 @@ def viewrecords(userid: UserID) -> Response:
     if info is None:
         abort(404)
     versions = {version: name for (game, version, name) in frontend.all_games()}
+    versions[10002] = 'MÚSECA PLUS'  # Hard code this version name for modpack songs
```
DragonMinded: This seems a bit suspect. I think I see what you're doing here, and it relates to the fact that MUSECA+ is not actually an onminix, but a fan continuation with new original songs. IIDX omnimix still made the assumption that every song would be found in SOME previous version, so there didn't need to be a separate category for it. This kinda makes the argument that this isn't really an omnimix so much as it is an expansion or a new mix entirely.

Granted, since its isolated to a dead game, and in the game's specific code, I think it's probably fine. However, that's the point of the `all_games()` method. I think you should be adding a hardcoded entry (and a comment as to why) here instead: https://github.com/DragonMinded/bemaniutils/blob/trunk/bemani/frontend/museca/museca.py#L25

I don't know why it was done in the spot it was for Pop'n.

bemani/utils/read.py
```diff
@@ -2839,15 +2839,20 @@ def __init__(
         no_combine: bool,
         update: bool,
     ) -> None:
-        actual_version = {
-            '1': VersionConstants.MUSECA,
-            '1+1/2': VersionConstants.MUSECA_1_PLUS,
-        }.get(version, -1)
+        if version in ['1', '1+1/2']:
```
DragonMinded, line 2842: Why can't this just be a new key in the above dictionary?

bemani/utils/read.py
```diff
@@ -2839,15 +2839,20 @@ def __init__(
...
+            actual_version = {
+                '1': VersionConstants.MUSECA,
+                '1+1/2': VersionConstants.MUSECA_1_PLUS,
+            }.get(version, -1)
+        elif version == 'plus':
+            actual_version = VersionConstants.MUSECA_1_PLUS + DBConstants.OMNIMIX_VERSION_BUMP
 
         if actual_version in [VersionConstants.MUSECA, VersionConstants.MUSECA_1_PLUS]:
             self.charts = [0, 1, 2, 3]
+        elif actual_version == VersionConstants.MUSECA_1_PLUS + DBConstants.OMNIMIX_VERSION_BUMP:
```
DragonMinded: Same here, unclear why it has to be a separate entry?

DragonMinded (commented):
Dang this was really clean overall, just a few comments.

.gitignore
Subject38: I just store all the private server files here
```
~/bemaniutils/server# ls
api.sock  bemani         frontend.wsgi  install  preload    scores_tsvs  services.sock  uwsgi
api.wsgi  frontend.sock  game_data      nginx    scheduler  server.yaml  services.wsgi
```

.gitignore
Subject38: :)

bemani/api/objects/catalog.py
Subject38: I plan to extend omnimix to popn, jubeat(already done on my main repository), and maybe even reflec beat. Should I just add every individual case as I add omnimix support? I feel like that just gets a bit overbearing but if it's preferred to explicitly say that there is no omnimix support for a game then I'll change it.

bemani/utils/read.py
Subject38: I wasn't sure how to handle it but will do

bemani/api/objects/catalog.py
DragonMinded: I prefer explicit, because it forces you to "turn on" support for omnimix in games. The reason is such: Say a remote server does support omnimix and this server doesn't. If we turn on omnimix support for all games, when a game requests scores, we will return an empty list. If we explicitly turn on omnimix per game, then we would at least return scores for the base game an omnimix is based off of.

bemani/frontend/popn/popn.py
```diff
@@ -33,6 +33,7 @@ class PopnMusicFrontend(FrontendBase):
 
     def all_games(self) -> Iterator[Tuple[str, int, str]]:
         yield from PopnMusicFactory.all_games()
+        yield ('pnm', 0, 'CS and Licenses')  # Folder that doesn't belong to any specific game
```
DragonMinded, line 36: Nice, fixing my own crap for me :>

DragonMinded (commented):
Looks good aside from the explicit omni enable/disable.

DragonMinded (approved):
I noticed a nit in the way you're yielding raw strings instead of game constants (not searchable) but its not worth it enough to hold up the review so I'll submit a fix after.

PR #12, opened 2021-01-24
Fix a little logic error in eacoin.py

kami-poi:
Hi,  
Nowaways I'm working on adding SDVX V support on your server. While reading your code, I found that it's a little logicial error on handle_eacoin_request method. Although it won't cause any bug while running, I also corrected it and  make a PR to show you there has some incorrect here.

### DIFF
--- a/bemani/backend/core/eacoin.py
+++ b/bemani/backend/core/eacoin.py
@@ -131,7 +131,7 @@ class PASELIHandler(Base):
                 return make_resp(2, 0)
 
             userid = self.data.local.user.from_session(session)
-            if session is None:
+            if userid is None:
                 # Refuse to do anything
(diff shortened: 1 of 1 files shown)


bemani/backend/core/eacoin.py
```diff
@@ -131,7 +131,7 @@ def make_resp(status: int, balance: int) -> Node:
                 return make_resp(2, 0)
 
             userid = self.data.local.user.from_session(session)
-            if session is None:
```
kami-poi, line 134: We have already checked nullable for session on L128. So I think we should check nullable for userid now.

DragonMinded: Ooh nice catch!

DragonMinded (approved):

PR #17, opened 2021-05-17
Support IIDX Cannon Ballers and Rootage

Subject38:


DragonMinded:
This is going nowhere, closing so it doesn't just sit here.

Subject38:
Uhhhhh. I wasn't really sure where to go from there??? Is there anything i was missing?

DragonMinded:
It was still in draft state, I thought you were going to make it non-draft when you were ready for it to review.

bemani/api/objects/catalog.py
```diff
@@ -155,11 +155,30 @@ def __format_jubeat_extras(self) -> Dict[str, List[Dict[str, Any]]]:
         else:
             return {"emblems": []}
 
+    def __format_iidx_extras(self) -> Dict[str, List[Dict[str, Any]]]:
```
DragonMinded, line 158: This should go along with an API documentation change in this section: https://github.com/DragonMinded/bemapi#iidx-additional-attributes-and-documentation-2

Much like SDVX has documentation for its own extra catalog entries.

bemani/api/objects/catalog.py
```diff
@@ -155,11 +155,30 @@ def __format_jubeat_extras(self) -> Dict[str, List[Dict[str, Any]]]:
...
+        # Gotta look up the unlock catalog
+        items = self.data.local.game.get_items(self.game, self.version)
+
+        return {
+            "qpros": [
+                {
+                    "identifier": item.data.get_str("identifier"),
```
DragonMinded, line 165: What's the difference between 'id' and 'identifier'? Those are the same word in English.

bemani/backend/iidx/cannonballers.py
```diff
@@ -1,15 +1,2184 @@
...
+        machine = self.data.local.machine.get_machine(self.config['machine']['pcbid'])
+        if machine.arcade is not None:
+            course = self.data.local.machine.get_settings(machine.arcade, self.game, self.music_version, 'shop_course')
+        else:
+            course = None
+
+        if course is None:
+            course = ValidatedDict()
+
+        root.set_attribute('music_0', str(course.get_int('music_0', 20032)))
```
DragonMinded, line 345: These are valid song IDs for songs found in CB, right? Just checking that they weren't copied without verifying.

bemani/backend/iidx/cannonballers.py
```diff
@@ -1,15 +1,2184 @@
...
+
+    # Bare minimum response to say we handled the request
+    def handle_IIDX25music_beginnerplay_request(self, request: Node) -> Node:
+        return Node.void('IIDX25music')
+
+    def handle_IIDX25music_arenaCPU_request(self, request: Node) -> Node:
+        root = Node.void('IIDX25music')
+        music_list = [0, 0, 0, 0]
+        cpu_list = [0, 0, 0, 0]
+        grade_to_percentage_map = {
```
DragonMinded, line 882: Where is this from?

bemani/backend/iidx/cannonballers.py
```diff
@@ -1,15 +1,2184 @@
...
+
+        newsong_another = Node.void('newsong_another')
+        root.add_child(newsong_another)
+        newsong_another.set_attribute('open', '1')
+
+        deller_bonus = Node.void('deller_bonus')
+        root.add_child(deller_bonus)
+        deller_bonus.set_attribute('open', '1')
+
+        common_evnet = Node.void('common_evnet')  # Yes, this is misspelled in the game
```
DragonMinded, line 981: Still?

bemani/backend/iidx/cannonballers.py
```diff
@@ -1,15 +1,2184 @@
...
+                'songs': [
+                    23082,
+                    21007,
+                    23037,
+                    25067,
+                ],
+            },
+        ]
+
+        # For some reason, omnimix crashes on course mode, so don't enable it
```
DragonMinded, line 1148: Is that still the case with CB omni? They seriously never fixed it?

bemani/backend/iidx/cannonballers.py
```diff
@@ -1,15 +1,2184 @@
...
+        now_combo = request.child_value('now_combo')
+        now_fast = request.child_value('now_fast')
+        now_slow = request.child_value('now_slow')
+        best_clear_string = clear_map.get(best_clear, 'NO PLAY')
+        now_clear_string = clear_map.get(now_clear, 'NO PLAY')
+        # let's get the song info first
+        song = self.data.local.music.get_song(self.game, self.music_version, music_id, class_id)
+        notecount = song.data.get('notecount', 0)
+        # Construct the dictionary for the broadcast
+        card_data = {
```
DragonMinded, line 1389: Ideally the left hand side of this dictionary should be constants, defined in the triggers class or common. Maybe for a future PR though? This looks brittle.

bemani/backend/iidx/rootage.py
```diff
@@ -0,0 +1,1949 @@
...
+                continue
+            if ach.id != rank:
+                continue
+            num_players = num_players + 1
+
+        root = Node.void('IIDX26grade')
+        root.set_attribute('pnum', str(num_players))
+        return root
+
+    def handle_IIDX26pc_common_request(self, request: Node) -> Node:
```
DragonMinded, line 815: I think you mentioned Rootage and above don't have courses anymore?

bemani/data/triggers.py
```diff
@@ -0,0 +1,37 @@
...
+            webhook = DiscordWebhook(url=self.config['webhooks']['discord'][game])
+            scoreembed = DiscordEmbed(title=f'New {game} Score!', color='fbba08')
+            scoreembed.set_footer(text=(now.strftime('Score was recorded on %m/%d/%y at %H:%M:%S')))
+
+            # lets give it an author
+            song_url = f"{self.config['server']['uri']}/{game}/topscores/{song.id}" if self.config['server']['uri'] is not None else None
+            scoreembed.set_author(name=self.config['name'], url=song_url)
+            for item in data:
+                inline = True
+                if item in ['DJ Name', 'Song', 'Artist', 'Play Stats']:
```
DragonMinded, line 33: This is where we would want to use those constants so we don't have magic strings floating around that are hard to grep for and brittle.

bemani/utils/psmap.py
```diff
@@ -67,9 +73,7 @@ def read_string(offset: int) -> str:
         except ValueError:
             pass
 
-        if nodetype == 0x00:
```
DragonMinded, line 70: Do we encounter node type 0 in 64-bit applications?

config/server.yaml
```diff
@@ -25,6 +25,12 @@ server:
     # Whether PCBIDs must be added to the network before games will work.
     enforce_pcbid: False
 
+# Webhook URLs
```
DragonMinded: This should probably be elaborated, like what this section enables, and that it can be left out entirely.

verifytraffic
```diff
@@ -14,6 +14,8 @@ declare -a arr=(
     "iidx-22"
     "iidx-23"
     "iidx-24"
+    "iidx-25"
```
DragonMinded, line 17: YES

verifytraffic
```diff
@@ -14,6 +14,8 @@ declare -a arr=(
...
+    "iidx-26"
```
DragonMinded, line 18: YES!

bemani/utils/read.py
```diff
@@ -1620,73 +1648,250 @@ def scrape(self, binfile: str, assets_dir: Optional[str]) -> List[Dict[str, Any]
...
             bh.close()
 
-        musicdb = IIDXMusicDB(binarydata)
+        import_qpros = True  # by default, try to import qpros
+        try:
+            pe = pefile.PE(data=binarydata, fast_load=True)
+        except BaseException:
+            import_qpros = False  # if it failed then we're reading a music db file, not the executable
+
+        def virtual_to_physical(offset: int) -> int:
```
DragonMinded, line 1657: This really should be moved to some common functionality, like common.pefile or something. We've copy-pasted it so many times by now...

bemani/utils/read.py
```diff
@@ -1620,73 +1648,250 @@ def scrape(self, binfile: str, assets_dir: Optional[str]) -> List[Dict[str, Any]
...
+                        'spn': notecounts[0],
+                        'sph': notecounts[1],
+                        'spa': notecounts[2],
+                        'dpn': notecounts[3],
+                        'dph': notecounts[4],
+                        'dpa': notecounts[5],
+                    },
+                })
+
+        qpros: List[Dict[str, Any]] = []
```
DragonMinded, line 1731: I should probably have formalized this, including some sort of check, but direct DLL reads are always vulnerable to the wrong datecode DLL being used. I have comments on the Pop'n stuff for exactly what datecode of data should be used in each, maybe we need that here as well?

DragonMinded (commented):
I assume that arena mode is only CPUs, since I don't see any lobby functionality for linking between actual cabinets? Aside from that this looks pretty good and I just need to pull this down and run it locally.

Not related to this PR, but I realize that read.py is getting extremely way out of hand and perhaps we should consider simplifying it in the future. We already detect file types in some instances, maybe we should drop --tsv/--csv/--xml/--bin options altogether and just deduce what is being imported by the file type? And we should probably do a best-effort check when reading data out of DLLs that the data is actually there and raise an exception with the expected datecode if it isn't? I dunno, its a mess.

bemani/backend/iidx/cannonballers.py
Subject38: i went and downloaded cb omnimix just to check this

DragonMinded:
I pulled this down and it all seems to work fine, so this is ready to merge as soon as review comments are responded to/addressed.

bemani/api/objects/catalog.py
Subject38: It's a stripped down filename that is essentially the name of the qpro but in ascii. Maybe ascii_name or just filename would be better but I've already been using this...

bemani/utils/psmap.py
Subject38: I removed this line because when I was using this to try to reverse some psmaps it would throw an exception. I forget the exact example that broke it.

bemani/api/objects/catalog.py
Subject38: Pull request submitted

bemani/backend/iidx/cannonballers.py
Subject38: Checked just now and they are in the musicdb

bemani/backend/iidx/cannonballers.py
Subject38: I pretty much just made up these values on the spot . Would it be better to make the constants on the left use the DAN_RANK constants? All this method exists to do is allow the CPUs to do something rather than just sit still

bemani/backend/iidx/cannonballers.py
Subject38: If there's anything i've learned working on konami games, they don't usually change old code or even get rid of it. The expert mode stuff is still in the net code for newer games and there's even a button in rootage even though the mode is completely removed

bemani/utils/read.py
Subject38: I simply forgot to add those comments so I'll add them shortly

bemani/backend/iidx/rootage.py
Subject38: Yes, interestingly ranking.getranker is still called but it's a no op pretty much now. Dan courses are of course still a thing but the expert mode stuff is all gone

bemani/api/objects/catalog.py
DragonMinded: As long as the documentation is updated (which is is) that's fine. It was not clear in the code and there was no comments.

bemani/backend/iidx/cannonballers.py
DragonMinded: Are they the DAN rank IDs on the left side? If so, then yes we should use those constants for documentation and such.

bemani/backend/iidx/cannonballers.py
DragonMinded: Can we at least put a TODO in here to do that work? I should probably have requested that...

bemani/utils/psmap.py
DragonMinded: Okay, but its needed then. Fair enough.

bemani/backend/iidx/cannonballers.py
DragonMinded: https://github.com/DragonMinded/bemaniutils/issues/27 I created a task so feel free to add a todo at your discretion.

bemani/utils/read.py
DragonMinded: https://github.com/DragonMinded/bemaniutils/issues/28 created for this, since its not applicable to this PR.

PR #20, opened 2021-06-07
Fix compatibility with python < 3.8 by switching from typing.Final to typing_extensions.Final

Subject38:


DragonMinded (approved):
LGTM. I checked it out and made sure it still typed on 3.8 as well as compiled with the experimental mypyc compiler and it's all perfect.

PR #22, opened 2021-06-28
api client: fix get_all_records

### LINKED ISSUE #21: Bemapi client: get_all_records does not merge records correctly
I ran into this bug because my personal instance is connected to 3 other instances. What would happen is the global record would be very obviously incorrect for some songs in sdvx. I managed to log the output of get_all_records and managed to confirm the cause to be the merging code.
image
image
As can be clearly seen here, there are two scores in the list for the same chart. The song that would be shown as the global record would be the former score since it shows up later in the list and would overwrite the earlier but better score. I'm raising this issue now because I'm currently working on a fix

Subject38:
Closes #21. Add a new method with slightly different behavior from __merge_global_scores and call that instead in get_all_records

### DIFF
--- a/bemani/data/api/music.py
+++ b/bemani/data/api/music.py
@@ -668,6 +668,67 @@ class GlobalMusicData(BaseGlobalData):
 
         return self.__merge_global_scores(game, version, localcards, localscores, remotescores)
 
+    def __merge_global_records(
+        self,
+        game: str,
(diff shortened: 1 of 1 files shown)


DragonMinded (approved):
Ya this looks good. Some tests would be nice, but the original code has none so whatever.

PR #25, opened 2021-07-06
frontend: Patch path traversal vulnerability

### LINKED ISSUE #23: Path Traversal in /jsx/:filename
https://github.com/DragonMinded/bemaniutils/blob/33d0000c771b6da22e45327921d1af21d9f83c81/bemani/frontend/app.py#L132-L137

Input to os.path.join isn't escaped - A parameter starting with `..%2F` will result in this endpoint sending back arbitrary files off of the system. The JSXTransformer call might incidentally protect against some path traversal by crashing, but it's likely this can be used maliciously.

Subject38:
Closes #23


DragonMinded:
Seems legit.

PR #31, opened 2021-08-17
iidx: Fix logic for saving miss count

Subject38:
It was recently brought to my attention that there's a logic error in saving miss_count. The current logic misses the edge case of failing to improve score but still lowering the miss_count. This new logic fixes this issue by rejecting -1 entirely.

### DIFF
--- a/bemani/backend/iidx/base.py
+++ b/bemani/backend/iidx/base.py
@@ -349,10 +349,11 @@ class IIDXBase(CoreHandler, CardManagerHandler, PASELIHandler, Base):
             # If it is a new score, create a new dictionary to add to
             scoredata = ValidatedDict({
                 'clear_status': clear_status,
-                'miss_count': miss_count,
                 'pgreats': pgreats,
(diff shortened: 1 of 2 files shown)


bemani/backend/iidx/base.py
```diff
@@ -364,8 +365,9 @@ def update_score(
             ex_score = max(ex_score, oldscore.points)
             scoredata = oldscore.data
             scoredata.replace_int('clear_status', max(scoredata.get_int('clear_status'), clear_status))
+            if miss_count != -1:
+                scoredata.replace_int('miss_count', min(scoredata.get_int('miss_count'), miss_count))
```
DragonMinded: It looks like this will still fail if a user already has a -1 in their miss count from a previous bug.

Subject38: Right thanks. Let me think up a solution

Subject38: Should be fixed now

PR #37, opened 2021-09-04
MGA / Pnm peace / Pnm score card

cracrayol:
* Added a basic support for Metal Gear Arcade
* Added the possibility to publish pop'n scorecard to Discord (start from éclale)
* Added better support for pop'n peace (phases data, updated read/trafficgen utility)


bemani/frontend/popn/endpoints.py
```diff
@@ -15,7 +15,7 @@
 popn_pages = Blueprint(
     'popn_pages',
     __name__,
-    url_prefix='/popn',
+    url_prefix='/pnm',
```
DragonMinded: This shouldn't matter, but curious why the change?

bemani/utils/frontend.py
```diff
@@ -1,5 +1,4 @@
 import argparse
-
```
DragonMinded: Repo coding style keeps a newline separating system/3rd party libs and imports from bemaniutils.

DragonMinded (commented):
Thanks for this! This mostly looks great, just need a client/traffic test for MGA. It looks mostly the same as Bishi, so it should be easy to capture a play session and create a client that emulates it for integration tests. Was Eclale the first game with eaappli in it? I think so. Also, I assume that any changes in peace for new events are not working yet?

I will probably refactor the way peace/usaneko share code to be similar to gravity wars with a usapeace.py that contains the stuff as an abstract base class and then usaneko.py and peace.py can both inherit from it, but I don't think you should need to do that work.

bemani/frontend/popn/endpoints.py
cracrayol: I've change this due to how song_url is made, in triggers.py. It takes the game.value as part of the url.

DragonMinded: That sounds legit enough to me. Maybe we should actually make the url prefix for each game the value of the enum? Seems like good tech debt cleanup for me or somebody else at some point.

cracrayol:
* Just need a client/traffic test for MGA -> done
* I assume that any changes in peace for new events are not working yet? -> For the phases of peace, i've used the dll from final data to retrieve the list of ids and the max value for each.
* I will probably refactor the way peace/usaneko share code to be similar to gravity wars with a usapeace.py -> From what I've seen, the current pnm iteration also use the same protocol than usaneko/peace but add some new achievements and account parameters.

bemani/backend/mga/mga.py
```diff
@@ -0,0 +1,177 @@
...
+
+        # Figure out what type of messsage this is
+        reqtype = data.child_value('datatype')
+        reqkey = data.child_value('datakey')
+
+        # System message
+        root = Node.void('system')
+
+        if reqtype == "S_SRVMSG" and reqkey == "INFO":
+            # Generate system message
```
DragonMinded, line 55: Could you elaborate on what this system message should enable? I'm a bit surprised they didn't keep the same style of unlock values as Bishi Bashi, but then again they like changing things for no reason.

bemani/backend/popn/usaneko.py
```diff
@@ -160,11 +160,12 @@ def __construct_common_info(self, root: Node) -> None:
             13: 1,
         }
 
-        for phaseid in phases:
+    def __construct_common_info(self, root: Node) -> None:
+        for phaseid in self.get_phases():
```
DragonMinded: Suggest `for phaseid, phase in self.get_phases().items():` here so that you don't have to call `self.get_phases()` repeatedly in the loop below. This will be come a bigger deal when event structures get hooked up to the arcade config screen.

bemani/client/mga/mga.py
```diff
@@ -0,0 +1,345 @@
+import base64
+import time
+from typing import Optional
+
+from bemani.client.base import BaseClient
+from bemani.protocol import Node
+
+
+class MetalGearArcadeClient(BaseClient):
```
DragonMinded, line 9: Nice, thanks for this! Could you also add it to `verifytraffic` so that MGA gets checked when somebody runs full integration tests?

bemani/data/triggers.py
```diff
@@ -33,7 +33,7 @@ def broadcast_score(self, data: Dict[BroadcastConstants, str], game: GameConstan
             self.broadcast_score_discord(data, game, song)
 
     def broadcast_score_discord(self, data: Dict[BroadcastConstants, str], game: GameConstants, song: Song) -> None:
-        if game == GameConstants.IIDX:
+        if game in [GameConstants.IIDX, GameConstants.POPN_MUSIC] :
```
DragonMinded: Not super big deal but suggest set notation instead of list notation since membership is `O(1)` in sets.

bemani/frontend/mga/endpoints.py
```diff
@@ -0,0 +1,175 @@
...
+    # the JS frontend to allow for Windows IME input of hiragana/katakana.
+    def conv(char: str) -> str:
+        i = ord(char)
+        if i >= 0xFF41 and i <= 0xFF5A:
+            return chr(i - (0xFF41 - 0xFF21))
+        else:
+            return char
+    name = ''.join([conv(a) for a in name])
+
+    if re.match(
```
DragonMinded, line 127: Is this verified or just copied from Bishi? If the latter, could you put a TODO here to verify the regex?

bemani/frontend/static/controllers/mga/settings.react.js
```diff
@@ -0,0 +1,242 @@
...
+                                    } else if(c == '='.charCodeAt(0)) {
+                                        c = 0xFF1D;
+                                    } else if(c == '>'.charCodeAt(0)) {
+                                        c = 0xFF1E;
+                                    } else if(c == '?'.charCodeAt(0)) {
+                                        c = 0xFF1F;
+                                    }
+                                    value = value + String.fromCharCode(c);
+                                }
+                                var nameRegex = new RegExp(
```
DragonMinded, line 76: Same comment here regarding verified or not.

DragonMinded (commented):
I checked this out locally to make sure it all looked good (and I wanted to see it!). The code is not lint-clean (`verifylint` will check this) and these should be fixed before checkin:

```
dragonminded@dragonminded-server:~/BemaniUtils$ ./verifylint
bemani/backend/mga/mga.py:4:1: F401 'typing.Any' imported but unused
bemani/backend/mga/mga.py:4:1: F401 'typing.Dict' imported but unused
bemani/backend/popn/base.py:293:121: E231 missing whitespace after ','
bemani/backend/popn/peace.py:74:10: W292 no newline at end of file
bemani/client/mga/mga.py:213:17: W291 trailing whitespace
bemani/client/mga/mga.py:238:45: W291 trailing whitespace
bemani/client/mga/mga.py:241:45: W291 trailing whitespace
bemani/client/mga/mga.py:244:45: W291 trailing whitespace
bemani/data/triggers.py:36:66: E203 whitespace before ':'
bemani/utils/read.py:971:1: W293 blank line contains whitespace
```

Similarly, there are some type errors:

```
dragonminded@dragonminded-server:~/BemaniUtils$ ./verifytyping
bemani/backend/popn/base.py:295: error: Argument 1 to "get_profile" of "Base" has incompatible type "int"; expected "UserID"
bemani/backend/popn/base.py:335: error: Argument 1 to "broadcast_score" of "Triggers" has incompatible type "Dict[BroadcastConstants, object]"; expected "Dict[BroadcastConstants, str]"
bemani/frontend/mga/mga.py:21: error: Return type "Iterator[Tuple[str, int, str]]" of "all_games" incompatible with return type "Iterator[Tuple[GameConstants, int, str]]" in supertype "FrontendBase"
bemani/frontend/mga/mga.py:22: error: Incompatible types in "yield from" (actual type "Tuple[GameConstants, int, str]", expected type "Tuple[str, int, str]")
bemani/backend/popn/eclale.py:440: error: Missing positional argument "stats" in call to "broadcast_score" of "PopnMusicBase"
bemani/backend/popn/eclale.py:440: error: Argument 6 to "broadcast_score" of "PopnMusicBase" has incompatible type "Dict[str, Any]"; expected "int"
bemani/backend/popn/usaneko.py:692: error: Missing positional argument "stats" in call to "broadcast_score" of "PopnMusicBase"
bemani/backend/popn/usaneko.py:692: error: Argument 6 to "broadcast_score" of "PopnMusicBase" has incompatible type "Dict[str, Any]"; expected "int"
Found 8 errors in 4 files (checked 40 source files)
```

bemani/backend/mga/mga.py
cracrayol: Sorry, I can't answer to this. In fact, I've just dump network communications between the game and another private network, who has made basic support for MGA. My main goal, when made this plugin, was to help a friend who has the cab but was bored to being not able to save progress and has to do the (very long) tutorial each time he wanted to play the game.

cracrayol:
I've fixed the lint errors but will try to fixes typing errors later, it's 2:30am.

DragonMinded:
No rush, whenever you can get to it is fine!

bemani/backend/mga/mga.py
DragonMinded: Okay, that's fine. Not sure if the values there mean anything since it doesn't match what I'm used to but somebody can fix it later.

DragonMinded:
Sorry, I just realized that my change to use GameConstants values for URLs conflicts with your change :(

DragonMinded:
Looks good to me, there's a few tiny nits but nothing worth holding back.

PR #45, opened 2022-02-16
sdvx: fix the bug when saving skill analyzer results.

RinChanNOWWW:
The backend always saves the new course result (`put_achievement`) and overwrites the old one. But the old result should not be overwritten if it is better.


DragonMinded:
Nice catch!

PR #47, opened 2022-02-19
Implement basic backend cache

### LINKED ISSUE #41: Pagination in database get_all_records and get_all_scores
In jubeat and a few other games (GFDM comes to mind) pagination is requested for the response. In jubeat, get_mdata is called 3 times and a valid response is to send 3 pages of records rather than 1 page of every record in the response. In the gfdm v games, the score request method only allows up to 250 entries per page, and keeps requesting until the server says it's done. This is a request for anyone willing to take it on to implement pagination in get_all_records and get_all_scores that grabs the first x number of entries by music ID. For example, grab all scores between music id 0 and 250, 250 and 500, etc...

Subject38:
Also paginate jubeat's get_mdata request. For now, this change is very experimental and could probably use some sort of testing. This is not ready to merge but review would be appreciated. The `cached` decorator is based heavily off the one from here: https://stackoverflow.com/a/16723844
If merged this will close #41 since paginating the database lookup will no longer be required if we can just do one big lookup first. I'm hesitant to add the new lib (pylibmc) to requirements.txt since this should likely be an optional thing to run since it also requires having memcached installed and running.

### DIFF
--- a/bemani/backend/jubeat/base.py
+++ b/bemani/backend/jubeat/base.py
@@ -64,7 +64,7 @@ class JubeatBase(CoreHandler, CardManagerHandler, PASELIHandler, Base):
         """
         return Node.void('gametop')
 
-    def format_scores(self, userid: UserID, profile: Profile, scores: List[Score]) -> Node:
+    def format_scores(self, userid: UserID, profile: Profile, scores: List[Score], mdata_ver: Optional[int]) -> Node:
         """
         Base handler for a score list. Given a userid, profile and a score list,
         return a Node representing a score list. Should be overridden.
@@ -133,7 +133,7 @@ class JubeatBase(CoreHandler, CardManagerHandler, PASELIHandler, Base):
 
         return self.format_profile(userid, profile)
 
-    def get_scores_by_extid(self, extid: Optional[int]) -> Optional[Node]:
+    def get_scores_by_extid(self, extid: Optional[int], mdata_ver: Optional[int]) -> Optional[Node]:
         """
         Given an ExtID, return a formatted score node. Similar rationale to
         get_profile_by_refid.
@@ -148,7 +148,7 @@ class JubeatBase(CoreHandler, CardManagerHandler, PASELIHandler, Base):
         profile = self.get_profile(userid)
         if profile is None:
             return None
-        return self.format_scores(userid, profile, scores)
+        return self.format_scores(userid, profile, scores, mdata_ver)
 
     def update_score(
         self,
--- a/bemani/backend/jubeat/clan.py
+++ b/bemani/backend/jubeat/clan.py
@@ -1038,18 +1038,8 @@ class JubeatClan(
         data = request.child('data')
         player = data.child('player')
         extid = player.child_value('jid')
-        mdata_ver = player.child_value('mdata_ver')  # Game requests mdata 3 times per profile for some reason
-        if mdata_ver != 1:
-            root = Node.void('gametop')
-            datanode = Node.void('data')
-            root.add_child(datanode)
-            player = Node.void('player')
-            datanode.add_child(player)
-            player.add_child(Node.s32('jid', extid))
-            playdata = Node.void('mdata_list')
-            player.add_child(playdata)
-            return root
-        root = self.get_scores_by_extid(extid)
+        mdata_ver = player.child_value('mdata_ver')
+        root = self.get_scores_by_extid(extid, mdata_ver)
         if root is None:
             root = Node.void('gametop')
             root.set_attribute('status', str(Status.NO_PROFILE))
@@ -1103,8 +1093,14 @@ class JubeatClan(
 
         return Node.void('gameend')
 
-    def format_scores(self, userid: UserID, profile: Profile, scores: List[Score]) -> Node:
-
+    def format_scores(self, userid: UserID, profile: Profile, scores: List[Score], mdata_ver: Optional[int]) -> Node:
+        if mdata_ver is None:
+            mdata_ver = 1
+        min_music_id, max_music_id = {
+            1: (0, 60000000),
+            2: (60000000, 90009999),
+            3: (90009999, 1000000000),
+        }.get(mdata_ver)
         root = Node.void('gametop')
         datanode = Node.void('data')
         root.add_child(datanode)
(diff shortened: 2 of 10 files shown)


bemani/backend/jubeat/base.py
```diff
@@ -64,7 +64,7 @@ def format_profile(self, userid: UserID, profile: Profile) -> Node:
         """
         return Node.void('gametop')
 
-    def format_scores(self, userid: UserID, profile: Profile, scores: List[Score]) -> Node:
+    def format_scores(self, userid: UserID, profile: Profile, scores: List[Score], mdata_ver: Optional[int]) -> Node:
```
DragonMinded, line 67: I would call this "page" or "part" or something more akin to what it does, instead of just what the game calls it. The game's name for this is about the worst name for something I've ever seen. Maybe even "chunk", not sure, but the current name doesn't make sense in any respect. Obviously the name should go in the other function prototypes as well.

bemani/backend/jubeat/clan.py
```diff
@@ -1103,8 +1093,14 @@ def handle_gameend_final_request(self, request: Node) -> Node:
...
 
-    def format_scores(self, userid: UserID, profile: Profile, scores: List[Score]) -> Node:
-
+    def format_scores(self, userid: UserID, profile: Profile, scores: List[Score], mdata_ver: Optional[int]) -> Node:
+        if mdata_ver is None:
+            mdata_ver = 1
+        min_music_id, max_music_id = {
+            1: (0, 60000000),
+            2: (60000000, 90009999),
+            3: (90009999, 1000000000),
```
DragonMinded, line 1102: This seems a bit arbitrary? Wouldn't that bias scores into different buckets depending on how new the game is?

bemani/backend/jubeat/clan.py
```diff
@@ -1164,6 +1160,8 @@ def format_scores(self, userid: UserID, profile: Profile, scores: List[Score]) -
                 music.replace_dict(str(score.id), data)
 
         for scoreid in music:
+            if int(scoreid) >= max_music_id or int(scoreid) <= min_music_id:
```
DragonMinded, line 1163: Having equals in both of these seems like a problem. Wouldn't that disregard the exact scoreid when comparing?

bemani/backend/jubeat/prop.py
```diff
@@ -536,17 +536,7 @@ def handle_gametop_get_mdata_request(self, request: Node) -> Node:
         player = data.child('player')
         extid = player.child_value('jid')
         mdata_ver = player.child_value('mdata_ver')  # Game requests mdata 3 times per profile for some reason
```
DragonMinded, line 538: Comment out of date now.

bemani/data/api/music.py
```diff
@@ -500,6 +501,7 @@ def get_scores_for_card(cardid: str) -> List[Score]:
 
         return topscore
 
+    @cached(lifetime=5)
```
DragonMinded, line 504: Is this 5 seconds? Hate that it doesn't specify units :\

bemani/data/cache.py
```diff
@@ -0,0 +1,53 @@
...
+    def _cached(func: Callable) -> Callable:
+        if has_mc:
+            memcache = pylibmc.Client(["127.0.0.1"], binary=True)
+            memcache.behaviors = {"tcp_nodelay": True, "ketama": True}
+        else:
+            memcache = None
+
+        @wraps(func)
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            if memcache is None:
```
DragonMinded, line 23: Do this outside the wrapper so it doesn't have to be checked every call. Just return a different constructed function that wraps without affecting anything.

bemani/data/cache.py
```diff
@@ -0,0 +1,53 @@
...
+try:
+    import pylibmc  # type: ignore
+    has_mc = True
+except ModuleNotFoundError:
+    has_mc = False
+
+
+def cached(lifetime: int=10, extra_key: Any=None) -> Callable:
+    def _cached(func: Callable) -> Callable:
+        if has_mc:
```
DragonMinded, line 15: This should be a configuration, not based on imports working.

bemani/data/cache.py
```diff
@@ -0,0 +1,53 @@
...
+            memcache = pylibmc.Client(["127.0.0.1"], binary=True)
```
DragonMinded, line 16: I'd suggest having this be a configuration value, and its existence allowing pylibmc to operate.

bemani/data/cache.py
```diff
@@ -0,0 +1,53 @@
...
+        @wraps(func)
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            if memcache is None:
+                return func(*args, **kwargs)
+
+            if lifetime is not None:
+
+                # Hash function args
+                items = kwargs.items()
+                hashable_args = (args[1:], sorted(list(items)))
```
DragonMinded, line 30: Why is the first argument left out here? This seems non-generic? Is this to skip self? If so, comment please.

DragonMinded (changes requested):
I don't see any configuration for this, just an implicit "if you install pylibmc and your memcached is on localhost it will work" clause. That's not discoverable or maintainable. It should be a config option and pylibmc should go into requirements.txt. If it can't be installed depending on the platform then its okay to have the conditional import, but add a platform specifier in requirements.

In general the logic looks good on this though, super excited!

bemani/backend/jubeat/clan.py
```diff
@@ -1103,8 +1093,14 @@ def handle_gameend_final_request(self, request: Node) -> Node:
 
         return Node.void('gameend')
 
-    def format_scores(self, userid: UserID, profile: Profile, scores: List[Score]) -> Node:
-
+    def format_scores(self, userid: UserID, profile: Profile, scores: List[Score], mdata_ver: Optional[int]) -> Node:
+        if mdata_ver is None:
+            mdata_ver = 1
```
DragonMinded, line 1098: Why is this optional then? Its confusing that the function would take an optional argument and if its not specified pick one instead of returning all music. I would default this to selecting from all music if mdata_ver is none here, not arbitrarily choosing the first one.

Subject38:
Closing until I come back and clean this up/switch to using redis instead for memcaching

PR #48, opened 2022-06-18
Implement decompressed kbinxml decoding

Bottersnike:
Implement decoding of decompressed binary packets.

See https://bsnk.me/eamuse/packet.html#schema for format reference.

## Potential todo:
- [ ] Switch to decompressed packing if any node names longer than 24h are present ('cus that's invalid!)

### DIFF
--- a/bemani/protocol/binary.py
+++ b/bemani/protocol/binary.py
@@ -483,7 +483,7 @@ class BinaryEncoder:
     A class capable of taking a Node tree and encoding it into a binary format.
     """
 
-    def __init__(self, tree: Node, encoding: str) -> None:
+    def __init__(self, tree: Node, encoding: str, compressed: bool=True) -> None:
         """
         Initialize the object.
 
@@ -498,6 +498,7 @@ class BinaryEncoder:
         self.__body: List[int] = []
         self.__body_len = 0
         self.executed = False
+        self.compressed = compressed
 
         # Generate the characer LUT
         self.char_lut: Dict[str, int] = {}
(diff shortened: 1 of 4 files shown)


DragonMinded:
For the TODO, would you do that by walking the tree recursively and choosing compressed/uncompressed if any node or attribute is > 0x24? Seems reasonable to do that as a transparent way of encoding without the above layers needing to know anything. I'd be cool if you had that in a different PR to keep this one simple. As for this one, do you have any sample binary data (maybe generated by another known good kbinxml implementation) that could be checked into the tests directory alongside the other fixture data and used to test this?

Bottersnike:
A pre-walk of the tree would likely be the plan yeah. I can look into generating some tests later today 👍

DragonMinded:
Pinging this, did you get a chance to write tests?

Bottersnike:
> Pinging this, did you get a chance to write tests?

Plenty, but never used any of them 😅. Will try and get some actually written

DragonMinded:
I know there's already some binary blobs from "difficult" bits I've had to fix the BINXML decoder on before, similarly for the LZSS encoder/decoder. If you just take a known working blob encoded by mon's tool or something and check it in and verify that it decodes, and then re-encodes back to the same blob, that'll probably be fine. I just want to make sure that it's not going to regress any existing decoding/encoding.

Bottersnike:
I've just pushed an update that also adds decompressed encoding and then uses it as part of the roundtrip loop, which should cover most interesting cases

DragonMinded:
Ah shit, I completely spaced on this, let me check now!

PR #51, opened 2022-08-16
Older IIDX routing refactor.

### LINKED ISSUE #50: Refactor older IIDX game classes to use the newer routing
IIDX 20-23 all use the older routing style where the method name is extracted by the game class rather than having the method in the function name. I was going to start working on a different refactor only to realize this issue existed. I'll be opening a pull request later tonight hopefully to get this done. This issue is just for tracking purposes

Subject38:
Fixes some bugs caught in traffic testing as well. Closes #50

### DIFF
--- a/bemani/backend/iidx/cannonballers.py
+++ b/bemani/backend/iidx/cannonballers.py
@@ -86,6 +86,14 @@ class IIDXCannonBallers(IIDXCourse, IIDXBase):
 
     FAVORITE_LIST_LENGTH: Final[int] = 20
 
+    GAME_CHART_TYPE_N7: Final[int] = 0
+    GAME_CHART_TYPE_H7: Final[int] = 1
+    GAME_CHART_TYPE_A7: Final[int] = 2
+    GAME_CHART_TYPE_N14: Final[int] = 3
+    GAME_CHART_TYPE_H14: Final[int] = 4
+    GAME_CHART_TYPE_A14: Final[int] = 5
+    GAME_CHART_TYPE_B7: Final[int] = 6
+
     requires_extended_regions: bool = True
 
     def previous_version(self) -> Optional[IIDXBase]:
@@ -290,6 +298,17 @@ class IIDXCannonBallers(IIDXCourse, IIDXBase):
         else:
             raise Exception('Invalid cltype!')
 
+    def game_to_db_chart(self, db_chart: int) -> int:
+        return {
+            self.GAME_CHART_TYPE_B7: self.CHART_TYPE_B7,
+            self.GAME_CHART_TYPE_N7: self.CHART_TYPE_N7,
+            self.GAME_CHART_TYPE_H7: self.CHART_TYPE_H7,
+            self.GAME_CHART_TYPE_A7: self.CHART_TYPE_A7,
+            self.GAME_CHART_TYPE_N14: self.CHART_TYPE_N14,
+            self.GAME_CHART_TYPE_H14: self.CHART_TYPE_H14,
+            self.GAME_CHART_TYPE_A14: self.CHART_TYPE_A14,
+        }[db_chart]
+
     def handle_IIDX25shop_getname_request(self, request: Node) -> Node:
         machine = self.data.local.machine.get_machine(self.config.machine.pcbid)
         if machine is not None:
@@ -367,7 +386,7 @@ class IIDXCannonBallers(IIDXCourse, IIDXBase):
 
     def handle_IIDX25ranking_getranker_request(self, request: Node) -> Node:
         root = Node.void('IIDX25ranking')
-        chart = int(request.attribute('clid'))
+        chart = self.game_to_db_chart(int(request.attribute('clid')))
         if chart not in [
             self.CHART_TYPE_N7,
             self.CHART_TYPE_H7,
@@ -461,7 +480,7 @@ class IIDXCannonBallers(IIDXCourse, IIDXBase):
     def handle_IIDX25ranking_entry_request(self, request: Node) -> Node:
         extid = int(request.attribute('iidxid'))
         courseid = int(request.attribute('coid'))
-        chart = int(request.attribute('clid'))
+        chart = self.game_to_db_chart(int(request.attribute('clid')))
         course_type = int(request.attribute('regist_type'))
         clear_status = self.game_to_db_status(int(request.attribute('clr')))
         pgreats = int(request.attribute('pgnum'))
@@ -597,7 +616,7 @@ class IIDXCannonBallers(IIDXCourse, IIDXBase):
 
     def handle_IIDX25music_appoint_request(self, request: Node) -> Node:
         musicid = int(request.attribute('mid'))
-        chart = int(request.attribute('clid'))
+        chart = self.game_to_db_chart(int(request.attribute('clid')))
         ghost_type = int(request.attribute('ctype'))
         extid = int(request.attribute('iidxid'))
         userid = self.data.remote.user.from_extid(self.game, self.version, extid)
@@ -673,7 +692,7 @@ class IIDXCannonBallers(IIDXCourse, IIDXBase):
     def handle_IIDX25music_reg_request(self, request: Node) -> Node:
         extid = int(request.attribute('iidxid'))
         musicid = int(request.attribute('mid'))
-        chart = int(request.attribute('clid'))
+        chart = self.game_to_db_chart(int(request.attribute('clid')))
         userid = self.data.remote.user.from_extid(self.game, self.version, extid)
 
         # See if we need to report global or shop scores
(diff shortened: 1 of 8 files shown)


Subject38:
All the refactoring work needed is now done. This wasn't strictly required for Heroic Verse and Bistrover to work, but it wouldn't make sense to have those games use the game_to_db_chart pattern and not do the same with the older games. I've been putting this stuff off for a while because it's just a bunch of copy paste busywork that needs to be done but certainly isn't fun or flashy lol. It does however make me thankful for the traffic tests that we have to have an easy benchmark to make sure nothing broke...

bemani/backend/iidx/cannonballers.py
```diff
@@ -290,6 +298,17 @@ def game_to_db_rank(self, game_dan: int, cltype: int) -> int:
         else:
             raise Exception('Invalid cltype!')
 
+    def game_to_db_chart(self, db_chart: int) -> int:
```
DragonMinded, line 301: No need for db_to_game_chart in 25?

bemani/backend/iidx/copula.py
```diff
@@ -288,306 +296,378 @@ def game_to_db_rank(self, game_dan: int, cltype: int) -> int:
...
+        return root
 
-        if method == 'getname':
-            root = Node.void('IIDX23shop')
-            root.set_attribute('cls_opt', '0')
-            machine = self.data.local.machine.get_machine(self.config.machine.pcbid)
-            root.set_attribute('opname', machine.name)
-            root.set_attribute('pid', str(self.get_machine_region()))
-            return root
+    def handle_IIDX23shop_savename_request(self, request: Node) -> Optional[Node]:
```
DragonMinded: Technically all of these can just be a `Node` return instead of an `Optional[Node]` return now that we split the handlers up. I don't think it should hold up the review tho unless you want to fix. I can go back and fix it myself if not.

bemani/backend/iidx/copula.py
```diff
@@ -599,788 +679,697 @@ def handle_IIDX23music_request(self, request: Node) -> Optional[Node]:
...
-            return root
+        return root
 
-        # Invalid method
-        return None
+    def handle_IIDX23grade_raised_request(self, request: Node) -> Optional[Node]:
+        extid = int(request.attribute('iidxid'))
+        cltype = int(request.attribute('gtype'))
+        rank = self.game_to_db_rank(int(request.attribute('gid')), cltype)
+        print(rank)
```
DragonMinded: Errant print statement spotted.

bemani/backend/iidx/pendual.py
```diff
@@ -414,6 +421,8 @@ def db_to_game_rank(self, db_dan: int, cltype: int) -> int:
             }[db_dan]
         elif cltype == self.GAME_CLTYPE_DOUBLE:
             return {
+                self.DAN_RANK_7_KYU: self.GAME_DP_DAN_RANK_5_KYU,
```
DragonMinded, line 424: How would this be possible, given that dan ranks don't span multiple games? Also, is it correct to map them all to 5kyu?

bemani/backend/iidx/pendual.py
```diff
@@ -601,749 +682,658 @@ def handle_IIDX22music_request(self, request: Node) -> Optional[Node]:
...
+        root.add_child(newsong_another)
+        newsong_another.set_attribute('open', '1')
+
+        limit = Node.void('limit')
+        root.add_child(limit)
+        limit.set_attribute('phase', '21')
+
+        # See if we configured event overrides
+        game_config = self.get_game_config()
+        if self.machine_joined_arcade():
```
DragonMinded, line 905: Why order change? Wouldn't get_game_config() run a DB operation that could be useless?

bemani/backend/iidx/spada.py
```diff
@@ -193,6 +201,8 @@ def db_to_game_rank(self, db_dan: int, cltype: int) -> int:
             }[db_dan]
         elif cltype == self.GAME_CLTYPE_DOUBLE:
             return {
+                self.DAN_RANK_7_KYU: self.GAME_DP_DAN_RANK_5_KYU,
```
DragonMinded, line 204: Same question here.

bemani/backend/iidx/tricoro.py
```diff
@@ -192,6 +200,8 @@ def db_to_game_rank(self, db_dan: int, cltype: int) -> int:
             }[db_dan]
         elif cltype == self.GAME_CLTYPE_DOUBLE:
             return {
+                self.DAN_RANK_7_KYU: self.GAME_DP_DAN_RANK_5_KYU,
```
DragonMinded, line 203: And here.

DragonMinded (commented):
Man what a cleanup. Is there anything left using the old routing method? I could give it a check. Also, I didn't see anywhere going from db->game chart. Is that because (IIRC) IIDX only ever sends data back in those stupid arrays?

bemani/backend/iidx/pendual.py
Subject38: Well, I ran into this specific issue because the server seems to pull profiles from later games??? That profile from a later game might be 6 or 7 kyu and throw an exception. I figured mapping to 5 kyu was fine but that's based on the assumption the game doesn't support anything below those for DP? Specifically in music_reg, `get_any_profiles` is called which if the user does not have a profile for the given version, will return a profile for the latest version

bemani/backend/iidx/pendual.py
Subject38: This was due to the original code being bugged. `game_config` is referred to later and might not be defined. I suppose a better way of doing it would be more like
```py
        # See if we configured event overrides
        if self.machine_joined_arcade():
            game_config = self.get_game_config()
            timeshift_override = game_config.get_int('cycle_config')
            ccxbm_on = game_config.get_bool('ccxbm_enable')
            omni_events = game_config.get_bool('omnimix_events_enabled')
            event_phase = game_config.get_int('event_phase')
        else:
            # If we aren't in an arcade, we turn off events
            timeshift_override = 0
            ccxbm_on = False
            omni_events = False
            event_phase = 0

        if self.omnimix and (not omni_events):
            phase = 0
        else:
            phase = event_phase
```

bemani/backend/iidx/cannonballers.py
Subject38: There's no `db_to_game_chart` because that is technically handled by `make_score_struct` in base.py. Unfortunately. `make_score_struct` will either likely need to be refactored or outright overloaded in heroic verse and later because the struct is changed. Currently I just overload the function but that feels wasteful...

bemani/backend/iidx/copula.py
Subject38: Oh thanks for catching this, I legitimately just forgot...

DragonMinded: NBD, but tighter typing always helps. In this case since not adding a return implicitly returns `None`, tightening the typing to `Node` instead of `Optional[Node]` will catch if you make a mistake and forget a return either in the default path or in some obscure early-return.

bemani/backend/iidx/cannonballers.py
DragonMinded: I need to have a look at that but I only have a few minutes so I'll peek and ponder later.

bemani/backend/iidx/pendual.py
DragonMinded: OH, the `get_any_profiles` code, right right. Okay yeah that's a fun edge case. That shouldn't be used much, but I think in some rival cases it comes up. Seems legit to keep a full mapping here then.

bemani/backend/iidx/pendual.py
DragonMinded: I agree that the proposed code looks better.

bemani/backend/iidx/cannonballers.py
DragonMinded: Ah, yeah. That and `make_beginner_struct` probably shouldn't go into base.py. That's my bad. That function is definitely specific to the parts of the series that use it, which admittedly is most of them. However, maybe they need to move to a common.py which subclasses from base.py or something, and then we can use it as a mixin to the older versions? Dunno, that's probably good enough and then do something different and custom for HV since its no longer a common function. Don't think any of that has to happen here though. Thoughts?

bemani/backend/iidx/pendual.py
```diff
@@ -911,14 +911,13 @@ def handle_IIDX22pc_common_request(self, request: Node) -> Node:
         else:
             # If we aren't in an arcade, we turn off events
             timeshift_override = 0
+            event_phase = 0
             ccxbm_on = False
             omni_events = False
 
         if self.omnimix and (not omni_events):
             phase = 0
-        elif self.omnimix and omni_events or not self.omnimix:
```
DragonMinded: This looks like it was correct before. Taking this line out means that omni_events doesn't actually do anything.

Subject38: If omnievents is set to True, then the else clause executes. If omnimix is not True, then the else clause will always execute. I'm not sure I see the issue...

DragonMinded: Oh my god I'm stupid you're completely right.

DragonMinded:
Let me do another pass on this tomorrow when I'm not completely stupid from exhaustion. I think this is ready 2 merge. Thanks again for helping with this cleanup!

DragonMinded (approved):
LGTM!

PR #53, opened 2022-08-16
Jubeat update

Subject38:
This was originally going to be a general housekeeping update. I realised about halfway through that a lot of this work is just about inseparable from my work on jubeat festo. This is by no means mergable but will serve as a sign of what I have so far. There's a lot of stuff that I'm unsure of how to proceed with.

- I have a better way of deriving the version a song came from than the method that just looks at the first digit of the songid, but I'm not sure how to migrate in such a way that the webui doesn't break entirely if the user doesn't reimport all the music_info.xml files for jubeat. 
- Similarly, I want to display the `detail_level` for 9s and 10s, but I don't know how to do so in a graceful manner that makes migrating from ints to floats for that item simple. Currently I just have it try and get the float and if that fails it gets the int instead but that seems like not a great idea. 
- Finally, I have a bit of a hacky way of getting the emblems imported into the webui assets but it's a nightmare that requires first importing the emblems to the correct location in the repository, and then installing the repository.

To be completely frank, this is an amount of work to get cleaned up that I'm not sure I have the will to do so. But it would be a shame for all of this work to only benefit my personal instance so I'm submitting this in good faith regardless.

Sorry for putting so much stuff to review in here. If I never get around to completing this, I hope some of this work can at least be helpful for your own efforts rather than starting from nothing.


DragonMinded:
Don't mind me, going to post questions about this on the relevant code.

bemani/utils/read.py
```diff
@@ -1344,6 +1344,27 @@ def scrape(self, xmlfile: str) -> Tuple[List[Dict[str, Any]], List[Dict[str, Any
             songid = int(music_entry.find('music_id').text)
             bpm_min = float(music_entry.find('bpm_min').text)
             bpm_max = float(music_entry.find('bpm_max').text)
+            earliest_version = 0
+            version = int(music_entry.find('version').text.strip(), 16)
```
DragonMinded: Interesting, so this is a bitfield, and its available on all supported game xmls? Wonder how I missed this.

DragonMinded: Wow, so it is....this is a good find! I think its fine to say "hey if you want fully correct versions, re-import all the XML files."

DragonMinded:
I don't see any update to the song TSV. Did you not have an update for that? That's usually my least favorite part of Jubeat. A few things in this looks like it can be cherrypicked directly into the trunk branch TBH. Not sure how you want to proceed if at all.

Subject38:
> I don't see any update to the song TSV. Did you not have an update for that? That's usually my least favorite part of Jubeat. A few things in this looks like it can be cherrypicked directly into the trunk branch TBH. Not sure how you want to proceed if at all.

I simply forgot to include it. You may notice a bunch of stuff not technically related to festo. The list has been amended a good amount for my own purposes because I used omnimix as well.

DragonMinded:
Appreciate it, definitely gives me a good starting point for festo stuff, do you want me to cherry-pick stuff and get it into master and deal with the merge issues yourself, or get smaller versions of this merged, or what? Basically, how do we get some of this rolled in?

Subject38:
Probably the best thing to do is cherry pick whatever you find useful for now. The webui work is super incomplete (it doesn't look at the imported version for a given song yet for example because the addition to read.py was a proof of concept). For stuff that can't be taken right now, we can work together on getting it cleaned up.

DragonMinded:
Speaking of jubeat, I'm looking at actually hooking up the recommend list by driving it off the most popular 12 songs for that version. Technically it should be able to look at who plays what in what groups and recommend based on what you did play other songs to play as well. However, that's a bit complicated for what I want to implement so I figure server best played is not a bad approximation.

DragonMinded:
Hmmm, it still needs to take into account the chart, so I guess it would have to be more clever than that.

DragonMinded:
Okay, got the version extraction rolled in, and I fixed the webui to use that instead of the division hack. Seems fine on my end for saucer-clan. Haven't gotten around to importing festo yet.

DragonMinded:
Got the inherit fix rolled in as well.

DragonMinded:
Rolled in part of the backend stuff (the stuff that doesn't have to do with your pagination changes), and rolled in the bemapi stuff. Also note that my update for the frontend also made bemapi expose the version as a 'category' much like other catalog objects when available.

DragonMinded:
Looked at the TSV stuff, the omnimix stuff looks fine to keep in mainline, so I rolled that in as well.

bemani/utils/read.py
```diff
@@ -1480,6 +1509,41 @@ def lookup(self, server: str, token: str) -> Tuple[List[Dict[str, Any]], List[Di
...
+            50000020: 80000037,  # 千本桜 was removed and then revived in clan
+            60000063: 70000100,  # Khamen break sdvx had the first id for prop(never released officially)
+        }
+
+        legacy_songid = legacy_to_modern_map.get(songid)
+        if legacy_songid is not None:
+            old_id = self.get_music_id_for_song(legacy_songid, chart)
+            if old_id is not None:
+                return old_id
+        modern_to_legacy_map = {
```
DragonMinded: I see no good reason not to create this by walking the `legacy_to_modern_map` in a comprehension and just swapping the key/value pair, like `{v, k for k, v in legacy_to_modern_map.items()}`.

bemani/utils/read.py
```diff
@@ -1499,22 +1563,39 @@ def import_music_db(self, songs: List[Dict[str, Any]]) -> None:
 
             self.start_batch()
             for chart in self.charts:
-                # First, try to find in the DB from another version
-                old_id = self.get_music_id_for_song(songid, chart)
-                if self.no_combine or old_id is None:
-                    # Insert original
-                    print(f"New entry for {songid} chart {chart}")
-                    next_id = self.get_next_music_id()
+                if(chart <= 2):
```
DragonMinded: I'm not quite sure I follow the logic here. Shouldn't there be something checking for whether the game is festo or above and not be inserting charts 3-5 for clan and below?

bemani/utils/read.py
```diff
@@ -1318,10 +1329,18 @@ def __init__(
...
             VersionConstants.JUBEAT_QUBELL,
             VersionConstants.JUBEAT_CLAN,
+            VersionConstants.JUBEAT_FESTO,
+            VersionConstants.JUBEAT_PROP + DBConstants.OMNIMIX_VERSION_BUMP,
+            VersionConstants.JUBEAT_QUBELL + DBConstants.OMNIMIX_VERSION_BUMP,
+            VersionConstants.JUBEAT_CLAN + DBConstants.OMNIMIX_VERSION_BUMP,
+            VersionConstants.JUBEAT_FESTO + DBConstants.OMNIMIX_VERSION_BUMP,
         ]:
-            self.charts = [0, 1, 2]
+            self.charts = [0, 1, 2, 3, 4, 5]
```
DragonMinded: Or would it be better to change the number of charts here? That is, after all, why this lookup exists. We should be able to return 0-2 for everything but festo, and 0-5 for festo, and then the below stuff works.

DragonMinded: Another advantage of doing it this way is we can do what IIDX does: gray out charts that are impossible for certain songs (such as songs that aren't in festo having grayed out hard scores).

DragonMinded: I went ahead and made that change and pushed the update.

DragonMinded:
Okay, pushed the updated code for read.py as well, including the small tweaks suggested.

DragonMinded:
I was able to pull a bunch of the stuff out of the frontend diff and finish up the rating changes as well as gray out charts that aren't supported, and also make the decimal rating system not activate when a song doesn't exist in festo or above. The hack for trying to grab a float and falling back to int seems to work fine by the way. What is software other than a collection of hacks?

DragonMinded:
Note that the frontend code and such hasn't been merged in since I am still working on a local fork. I can push the partial update branch though if you want to work off of it or merge stuff back.

DragonMinded:
Its here if you want to compare. I haven't merged in your changes yet aside from fixing up the frontend based on your code plus some cleanup/fixes. https://github.com/DragonMinded/bemaniutils/tree/jubeat_festo

Subject38:
I see you included Omnimix changes but not the change to using self.music_version. Is this intentional? Since importing omnimix will import at that game's version + 10000, I imagine without all the music_version stuff, it won't work...

DragonMinded:
No, I missed that :(

I am exhausted, sorry.

DragonMinded:
Okay, I rolled the omni stuff in, as well as one missed spot in common.py. I think outside of the unfinished web stuff (and import script) and your changes to mdata_ver that its all integrated into trunk. I haven't gotten the chance to really dig into the festo stuff yet so I haven't pulled your changes from that either, but the frontend stuff is pulled into the festo branch and I'm gonna keep pounding away at it.

DragonMinded:
As much as I can has been merged from the festo support here into the jubeat_festo branch. I still need to browse bemaniwiki and figure out what courses are missing and include those, and also go through and test a bunch of stuff and write traffic tests.

DragonMinded:
Very slow progress, but I got all the courses supported. I hit a snag supporting the random 10s course: Enabling it means it is actually the 61st course and the game has a hard limit of 60 courses. I guess they never hit this in production because they didn't keep a lot of the time limited courses around. So I "fixed" it by making the frontend have a KAC phase so you can choose which KAC courses are available (none, 8th, 9th, 10th). Would have rather just had all courses available but that's not possible.

I have to go back and verify a few other odds and ends and then I'll make traffic tests for festo and promote it to fully working. Did I miss anything else from this PR? If not, I think its safe to close soon.

DragonMinded:
Also, after I'm done with this I think I'm going to go back and try to add lobby support for Jubeat series. Feels like its about time.

Subject38:
Seems a lot of webui updates are still missing... Other than that I don't see a ton of issues? I changed the base to your jubeat_festo branch to see the difference.

DragonMinded:
Oh good call, let me see what all I'm missing still.

DragonMinded:
I'll have to figure out what's up, because several of the web changes I thought I did...

DragonMinded:
Okay, on spot check, a LOT of the files that it is showing different on top of the jubeat_festo branch have the changes upon checking. I don't know why the diff tool is showing so much that is already there. Is it screwing up and using the old base because of merge conflicts?

Subject38:
I'll rebase the branch later then. my first glance shows the jubility table isn't displayed yet though

DragonMinded:
Yeah, the jubility frontend definitely didn't make it. I noticed that the game never gets sent back the jubility stuff that gets saved, so I assume you were doing that just for the frontend. I was trying to keep to core implementation for base festo support because otherwise I was getting overwhelmed :<

DragonMinded:
I think, if I can get the basics tested (including traffic tests) and it doesn't make any major mistakes, and you rebase this on top of that work, then this can turn into a much simpler and more tractable PR for adding the extra features.

DragonMinded:
Okay this is rebased on top of the festo branch, and outside of changing the fill in category round tripping it looks to be functionally identical in terms of the core, so I'm going to merge the jubeat_festo branch and then we can get the emblem and stats updates in properly!

DragonMinded:
WTF I did not close this?! I think github closed it automatically when I deleted the base branch.

PR #54, opened 2022-09-04
Iidx hv and bistro

Subject38:
Missing traffic tests. I can't think of any obvious issues with the pull request in its current state though so I'm submitting as is for now. I'll do traffic tests in a bit (Already recorded packet logs for them I just need to sit down and do the rest). If you have any design objections feel free to bring them up though! :) As an aside, it might be advisable to update the bemapi spec to account for the new iidx charts...


bemani/format/afp/render.py
```diff
@@ -1099,7 +1099,7 @@ def __place(self, tag: Tag, operating_clip: PlacedClip, prefix: str = "") -> Tup
                 # Get rid of the objects that we're removing from the master list.
                 operating_clip.placed_objects = [
                     obj for obj in operating_clip.placed_objects
-                    if not(obj.object_id == tag.object_id and obj.depth == tag.depth)
+                    if not (obj.object_id == tag.object_id and obj.depth == tag.depth)
```
DragonMinded, line 1102: I was like "woah somebody else is working on the AFP renderer?!" and then I was like ....oh.

bemani/format/iidxmusicdb.py
```diff
@@ -101,29 +107,51 @@ def parse_string(string: bytes) -> str:
...
+                    parse_string(songdata[0]),
+                    parse_string(songdata[1]),
+                    parse_string(songdata[2]),
+                    parse_string(songdata[3]),
+                    [songdata[6], songdata[7], songdata[8], songdata[11], songdata[12], songdata[13], songdata[5], songdata[9], songdata[10], songdata[14]],
+                    songdata[4],
+                )
             if song.artist == 'event_data' and song.genre == 'event_data':
                 continue
-            self.__songs[songdata[11]] = (song, songoffset)
```
DragonMinded, line 126: Oh my god this code is horrible LMAO what was I thinking when I wrote this?!

bemani/frontend/iidx/iidx.py
```diff
@@ -72,6 +76,111 @@ def get_duplicate_id(self, musicid: int, chart: int) -> Optional[Tuple[int, int]
             16102: 21258,
             16101: 21262,
             14100: 21220,
+            28100: 27115,
+            # Tons of leggendarias here...
```
DragonMinded, line 80: Oh my god what a pain. Thank you for figuring this out.

bemani/utils/read.py
```diff
@@ -1752,6 +1761,7 @@ def __revivals(self, songid: int, chart: int) -> Optional[int]:
             21258: 16102,
             21262: 16101,
             21220: 14100,
+            27115: 28100,  # Fucking Fire Beat
```
DragonMinded, line 1764: Funny, that's what I say whenever I'm trying to do hacks for KBM/BMIII/etc...

bemani/utils/read.py
```diff
@@ -1811,6 +1785,79 @@ def __revivals(self, songid: int, chart: int) -> Optional[int]:
             if old_id is not None:
                 return old_id
 
+legg_map = {
+    1017: 1100, 4005: 4100,
```
DragonMinded, line 1789: The double-things-on-line threw me, but I don't think its worth fixing. Honestly we really should just say "We are formatted with black", do a bite-the-bullet pass and move on from there.

bemani/backend/iidx/heroicverse.py
```diff
@@ -1,17 +1,2029 @@
...
+        play_stats = self.get_play_statistics(userid)
+
+        # Look up judge window adjustments
+        judge_dict = profile.get_dict('machine_judge_adjust')
+        machine_judge = judge_dict.get_dict(self.config.machine.pcbid)
+
+        # Profile data
+        pcdata = Node.void('pcdata')
+        root.add_child(pcdata)
+        pcdata.set_attribute('id', str(profile.extid))
```
DragonMinded, line 1171: Sure hope you got to generate this automatically...

DragonMinded (commented):
General direction looks fine to me, I didn't see anything that would be a dealbreaker. There's only CPU arena in this, right? No player-to-player arena? Is that because of missing lobby functionality or is there something more complicated that needs to be done? Don't forget to update the readme with the additionally supported games. Thanks for tackling this by the way!

Subject38:
No online arena, but local arena should work by default. I personally don't find enough of a use for online lobby to bother implementing it for any game... With games like iidx, I have to wonder if it might be more advisable to make a separate project just for scraping the musicdbs with charts, and then just having some json blobs stored in this repository with which to seed the database. Because of how much manual mangling has to be done, I just have to wonder if using the chart parser to create a unique id for each chart and using that to determine the server id would be better than what we currently do. For example, with the modern to legacy stuff, I'm honestly not 100% certain the list is completely accurate, and checking for compete accuracy would be a pretty painstaking process if done manually...

DragonMinded:
Having a repo that is good at one thing: scraping metadata from ANY game ANYWHERE would actually be a pretty useful thing to have. If its structured right, bemaniutils can just depend on it and have a simple wrapper or something. I have, however, shied away from any scraped or copyright data in this repo, hence why you need to seed your installation from another active one or from your own sourced game files. Wouldn't be against somebody taking the read code, putting it in its own repo, cleaning it up and also checking in per-game blobs. I just won't be the one to do that ;)

bemani/utils/read.py
```diff
@@ -1811,6 +1785,79 @@ def __revivals(self, songid: int, chart: int) -> Optional[int]:
...
+    16030: 21252, 14012: 21264,
+    16034: 21254, 16038: 21255,
+    16040: 21256, 16042: 21257,
+    16047: 21259, 16049: 21260,
+    15005: 21261, 15008: 21263,
+    22008: 22101, 22013: 22102,
+    22024: 22103, 22027: 22104,
+    22031: 22105, 22089: 22106,
+    22006: 22107,
+}
```
ishmael573, line 1841: This bit is not properly indented.

data/iidx.tsv
```diff
@@ -26,3 +26,10 @@
 26037		CANVAS feat. Quimär	
 26070	Xlø		
 26101	Geirskögul		
+27083		DÉ DÉ MOUSE	
+27004	Dans la nuit de l'eternite		
+27035		CANVAS feat. Quimär	
+28007		CANVAS feat. Quimär & BEMANI Sound Team "L.E.D.-G"	
+28017	Pārvatī		
+28050	POLꞰAMAИIA		
+28084	Ignis†Iræ
```
ishmael573, line 35: Another tab needs to be inserted at the end of the line.

ishmael573 (changes requested):

data/iidx.tsv
Subject38: Gonna be honest, I underestimated just how much my ability to work on this would drop once I started working a "real" job. Thanks for the notes though. If you wanted to take over the pull request you would have my thanks tbh.

ishmael573:
I've been running this branch for a little while. It seems okay for now, but I've spotted few issues within the code. Suggestions above are what I've found.

Also, `bemani/frontend/static/iidx-options.js` file needs to be updated with options for 27 and 28.

Subject38:
replaced by #58

PR #55, opened 2022-09-18
Add Pnm Kaimei riddles support

cracrayol:
Add support for Pop'n Music Kaimei Riddles.
It's mainly based on the same code as peace, but with more (documented) params and additional code in (un)format_profile in order to manage the Kaimei! MN tanteisha event and new values in player profile.

### DIFF
--- a/bemani/backend/popn/kaimei.py
+++ b/bemani/backend/popn/kaimei.py
@@ -1,0 +1,491 @@
+# vim: set fileencoding=utf-8
+import math
+import random
+from typing import Any, Dict, List, Tuple
+
+from bemani.backend.popn.base import PopnMusicBase
+from bemani.backend.popn.common import PopnMusicModernBase
+from bemani.backend.popn.peace import PopnMusicPeace
+from bemani.common import VersionConstants
+from bemani.common.validateddict import Profile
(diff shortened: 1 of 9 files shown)


bemani/backend/popn/kaimei.py
```diff
@@ -0,0 +1,491 @@
...
+                # 3: 1982, 1983, 1988
+                # 4: 1986, 1987, 1993
+                # 5: 2017
+                30: popn_quest_lively_2,
+            },
+            False,
+        )
+
+    def format_profile(self, userid: UserID, profile: Profile) -> Node:
+        root = PopnMusicModernBase.format_profile(self, userid, profile)
```
DragonMinded, line 385: Technically this should use the magic `super()` instead of hand-specifying the base class, but given there's full traffic tests its trivial for somebody to fix it so I don't think it's a big deal.

DragonMinded (approved):
Nice, thanks! Looks complete (including traffic tests), let me pull this down sometime soon and double-check everything locally and then I'll merge.

DragonMinded:
Tested this all out locally, seems to work perfectly. Thanks again!

PR #57, opened 2022-09-23
add gitadora support.

wz18207:
~~client file havent finished. ill complete them later.~~
~~others fine.~~
edit. whole structs seems done. feel free to check and ask. if some places are confused.
so far the way~ we wait for the day~


DragonMinded:
First off, thanks! I had a quick look and one thing that stood out is you have a file that contains the music for both normal and omnimix for the game you are supporting. Music info should be read from the music DB that's populated by read.py, not by including a file that's been generated from a game's music. I see you are even scraping the database, why are you not using that instead of constants?

bemani/backend/gitadora/nextage.py
```diff
@@ -0,0 +1,1568 @@
+# vim: set fileencoding=utf-8
+import copy
+from typing_extensions import Final
+from typing import Optional, List, Dict
+
+from bemani.backend.gitadora.base import GitadoraBase
+from bemani.backend.gitadora.gitadoramusiclist import MUSICLIST_NEXTAGE, MUSICLIST_NEXTAGE_OMNIMIX
```
DragonMinded: This is what I refer to, there shouldn't be hardcoded music lists in the code.

bemani/backend/gitadora/nextage.py
```diff
@@ -0,0 +1,1568 @@
...
+        hot.add_child(Node.s32('major',1))
+        hot.add_child(Node.s32('minor',0))
+
+        # musicinfo. should been import the musicdb at first.
+        # gitadora's music difficulties have changed a lot between each version. 
+        # thus have to collect full music difficulties from each data's musiclist.
+        if self.omnimix == True:
+            all_songs = MUSICLIST_NEXTAGE_OMNIMIX
+        else:
+            all_songs = MUSICLIST_NEXTAGE
```
DragonMinded, line 809: Not sure why these aren't loaded from data.local.music.

bemani/utils/read.py
```diff
@@ -4103,5 +4316,24 @@ def import_music_db(self, songs: List[Dict[str, Any]]) -> None:
         danevo.import_music_db(songs)
         danevo.close()
 
+    elif series == GameConstants.GITADORA:
+        gitadora = ImportGitadora(config, args.version, args.no_combine, args.update)
+        if args.tsv is not None:
```
DragonMinded, line 4489: I don't see a tsv file included in this diff.

wz18207:
> First off, thanks! I had a quick look and one thing that stood out is you have a file that contains the music for both normal and omnimix for the game you are supporting. Music info should be read from the music DB that's populated by read.py, not by including a file that's been generated from a game's music. I see you are even scraping the database, why are you not using that instead of constants?

Because the music skills in gitadora are extremely dependent on the difficulties of the songs, but each generation changes the difficulty of the songs, and the rating changes greatly, and the number of changes is large, which may affect the rating of the player's music skills.  So can't get data from old song difficulty
e.g.: 
https://remywiki.com/Kiwi
https://remywiki.com/Sakasama_Cinderella_parade
etc.

bemani/utils/read.py
wz18207: oops. i forget add this. ehh i would add tsv later. sorry~

bemani/backend/gitadora/nextage.py
wz18207: because musicdb doesnt include all hot music ids of each version. i have to grab  the hot music from bwiki for each one. also its the game gf_skills and dm_skills issues.

wz18207:
http://bemaniwiki.com/?GITADORA+HIGH-VOLTAGE/%B5%EC%B6%CA%A5%EA%A5%B9%A5%C8%28GITADORA%A1%C1NEX%2BAGE%29
too much changes. i have to make a new songs diffifulties list for the game-send. this will make sure that players could get their correct playskills for each game versions.
Edit: high voltage is an example. add files contains full musiclist in order to add gitadora games support earlier than nextage uin the future. e.g.: exchain, matrixx, etc. these versions have the same status from nextage -> high voltage.

bemani/backend/gitadora/nextage.py
DragonMinded: How much of the info is not present in the game's music DB? Trying to get a handle on what the issue is. Nothing else in the network requires a music list like this so its suspicious (and I don't want to import the code as-is). Some other games like Jubeat are missing some info so they include a TSV file for augmenting the game data. Can this be done similar?

DragonMinded:
I'm not sure I understand. The music DB is game and version specific, and only ties the same song across multiple versions by using a common music ID. Each time you call "get_all_songs" it returns only the data scraped for that particular game/version. Many games also change information from game to game including game-specific IDs and the like.

bemani/backend/gitadora/nextage.py
wz18207: hot music items for each version of gitadora.  jubeat tsv is used to replace correct names and artists that are not shown in musicdb.xml. But hot music items is different for each version of gitadora, also hot music affects the player's skill score

wz18207:
yeah. thats true in gitadora.

DragonMinded:
I am still not sure why that means gitadora needs hardcoded music DB though...

bemani/backend/gitadora/nextage.py
DragonMinded: The musiclist looks like a complete list though, not just a list of song IDs included in the hot list.

wz18207:
the difficulty changes. gitadora total player skills(gf and dm) are highly based on songs difficulties. however songs difficulties level have changed too much from two versions. use hard code musicdb would send 'correct' music level and skills based on official eamuse. also doing this way would help me to make some lower gitadora versions. e.g. exchain, matrixx onedrive etc.

DragonMinded:
Are those difficulty levels not inside the game's music files? Where did you get those difficulties?

wz18207:
Because the difficulty of the original song id and difficulties will be inherited when read.py is read, so I have to do this.

DragonMinded:
Song difficulty will only be overwritten if its the same game/version. You should be using different version numbers for different released versions, at which point the difficulties will not be overwritten. That's the whole point of importing per-version and using the game/version song database within the game's backend and frontend. Where did you get the data that's in the hardcoded music list? If it came from the game's files, then it should be imported in read.py and read through get_all_songs().

wz18207:
full music list i grab from musicdb inside the game.
all hot musics i grab from bemaniwiki.
emm. ill edit my code later

DragonMinded (commented):
I pulled this down and I'm checking it now. I had to make a few typing fixes and I needed to rename "React.createClass" to "createReactClass" to make it compatible with the new frontend. Also put the drop-down in the right order alphabetically.

A few things I've noticed:
 - I do not see any bit in the readme about support or how to import data from the game.
 - There seems to be no game records page link. I'll add that to the code. Seems to work though.
 - The records display is a bit messy with how many charts there are. I wonder if we should eventually split up by guitar/drum/bass using tabs?

Once I verify the traffic tests are working I think this is good enough to merge and can be worked on more.

DragonMinded:
I am not sure which XML files to import, please document how you are supposed to bootstrap this.

DragonMinded:
I've figured out how to import now, but it looks like the traffic test client is unfinished, so that is a blocker before this can be merged.

DragonMinded:
For clarity, when I run the traffic tests I get the following. If the traffic tests are complete and working I think this can be checked in.

```
dragonminded@dragonminded-server:~/BemaniUtils$ ./trafficgen --game gitadora-7
Emulating Gitadora Nextage
Traceback (most recent call last):
  File "./trafficgen", line 11, in <module>
    runpy.run_module(f"bemani.utils.{name}", run_name="__main__")
  File "/usr/lib/python3.8/runpy.py", line 210, in run_module
    return _run_code(code, {}, init_globals, run_name, mod_spec)
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/dragonminded/BemaniUtils/bemani/utils/trafficgen.py", line 640, in <module>
    main()
  File "/home/dragonminded/BemaniUtils/bemani/utils/trafficgen.py", line 636, in main
    mainloop(args.address, args.port, args.config, action, game, args.cardid, args.verbose)
  File "/home/dragonminded/BemaniUtils/bemani/utils/trafficgen.py", line 570, in mainloop
    emu.verify(cardid)
  File "/home/dragonminded/BemaniUtils/bemani/client/base.py", line 424, in verify
    raise Exception('Override in subclass!')
Exception: Override in subclass!
```

wz18207:
emm, let me check my code. seems client broken. however all backend its fine.
To be honest I'm not sure how to write client file, but backend and frontend have been tested.

DragonMinded:
You can capture logs of the back-and-forth between the game and server and then the client emulates different packets to ensure that the server replies with the correct shape and such, you can look at the other clients for examples. For big complicated packets you can use `responsegen` to autogenerate code that writes packets in some cases. You will need to test the general startup, global server packets, creating, pulling and updating profile and also submitting and loading scores.

If the tests do not exist then there is no way to regression test that things have not broken and I can't take code that requires manual validation every time something changes.

wz18207:
I've been a bit busy with my hands lately and couldn't get away. If I have enough free time I will try to fix the problem as much as possible, sorry for the delay ._.

wz18207:
happy christmas and new year. (This sounds like it's a bit late lol). im so sorry with the delay. too busy at these time. i tried my best to finish the client files. however i have no idea that how i can test it.

wz18207:
Temp i have to close this pull request because there are many difference between the main currently.
i would start a new pull request for gitadora.

PR #58, opened 2022-09-24
Support IIDX 27 and 28

ishmael573:
This is an extension of #54.

Original commits are cherry-picked as-is, and my set of patches are applied as well.

Some are already commented by myself on the original PR, some are new.

---

To-dos:

- duplicated song id 23070 on `frontend/iidx/iidx.py`
- missing options on `bemani/frontend/static/iidx-options.js`


ishmael573:
Rebased and force-pushed.

I've also added game options for 28. Most of the options are filled, but `frame` and `bgm` are missing. Some are just copy and paste from previous version.

27's options are missing because I don't have data at the moment. But it might be trivial with `bm2dx.dll` and some IDA fiddling.

Furthermore, duplicate options between `towel` and `pacemaker` are removed.

ishmael573:
Rebased and force-pushed again.

This time, all the Black format changes made difficult to keep original two commits, so I've squashed few commits into one.  
Plus, HEROIC VERSE options are added, but 27 is n-3 now. Who cares?

Quite late, but okay to go, I suppose.

DragonMinded:
Hey, sorry about the conflicting files, but I tried at least in the music DB one to grab what you'd done as a baseline when I fixed it up to be more generic. For read, it's probably clashing with the fact that read.py now supports auto-determining the DLL version based on a set of offsets instead of one offset, which probably is messing up the qpro stuff. Also, I don't see any tests at all whatsoever. I can't take anything in that doesn't have comprehensive traffic tests, because then I can't regression test anything when I make changes.

PR #63, opened 2022-10-22
[sdvx]6代新组段位+段位称号组BUG修正

SimoTsuKi:
[bemani.zip](https://github.com/DragonMinded/bemaniutils/files/9843589/bemani.zip)

PR #64, opened 2022-10-25
Support SDVX 06 (EXCEED GEAR)

ishmael573:
Same as #63, but better, I suppose.

A while ago, I've discovered that fork and that branch to be used on my personal instance. But it was full of unrelated changes, mainly PostgreSQL related. I've condensed them into one single commit with only necessary and minimal changes.

Notes:

- `bemani/backend/sdvx/exceedgear.py` is slightly changed from that branch and reordered here and there to minimize diff between `heavenlyhaven.py`.
- Use of `use_information` flag is commented out. I think this option is for displaying custom message on entry screen, but I've not tested this. These strings will probably use same format as `data/others/gameover.template.xml`. If this feature is being implemented correctly, two more `strs` and `longstrs` game setting entry might be needed.
- `use_asphyxia_gameover` flag is also commented out, but this thing is unknown, as original code suggests.
- Original code has implemented Discord broadcasting support in it, but I've commented it out for now.
- I think `force_unlock_crew` is not working properly?

### DIFF
--- a/bemani/api/app.py
+++ b/bemani/api/app.py
@@ -289,6 +289,7 @@ def lookup(protoversion: str, requestgame: str, requestversion: str) -> Dict[str
                 "2": VersionConstants.SDVX_INFINITE_INFECTION,
                 "3": VersionConstants.SDVX_GRAVITY_WARS,
                 "4": VersionConstants.SDVX_HEAVENLY_HAVEN,
+                "6": VersionConstants.SDVX_EXCEED_GEAR,
             },
         }
         .get(game, {})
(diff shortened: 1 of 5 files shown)


Subject38:
the issue with force_unlock_crew iirc is that the type for crew member should be 11, not 4

DragonMinded:
I appreciate the effort and work! However, bemaniutils strives to be a preservation tool, used for playing older games that are no longer available. We do not wish to compete with Konami on any current game that is out. That doesn't fit the bill of preserving old game experiences and directly competes with Konami's interests which is kinda a big no-no. We can revisit any sort of support once SDVX 6 is end-of-life and no longer making money for them.

PR #66, opened 2022-11-26
Refactor and better document card cipher

nullableVoidPtr:
This of course could all be simplified to use `Crypto.Cipher.DES3` but I can't figure out what was the original master key (24 bytes) that was used from the pre-computed expanded key.


DragonMinded:
Oh wow, thanks! How did I miss this? I assume this works for V1 and V2 cards (the original gray and the suica compatible) with th old 16-digit card backs and that it passes the unit tests for the cipher?

nullableVoidPtr:
I have since found a [source](https://bsnk.me/eamuse/cardid.html) which both properly reversed the cipher and identified the original 3DES key, and amended the PR with the simplified code (without the DES implementation).

The relevant unit test passed, but I don't have any "real data" or actual cards on hand to test this. I'm only going off of the unit test, which is obviously not an entire test coverage.

DragonMinded:
Hey I completely forgot about this again. I'm exporting all the cards from my local instance, running them through the cypher both ways, and then going to compare that to post-patching this. They should, in theory, match 100%, which will tell me all I need to know.

DragonMinded:
Everything matches to the byte. I think this is good.

PR #69, opened 2023-02-25
Add support for Pop'n music Omnimix v2

xinrin:


DragonMinded:
Thanks for the commit! I haven't looked at Pop'n omnimix, so I assume that it has XML files that it comes with for the songs that should be applied to it? Hence the loading of song info from XML. Also, this shouldn't load the data directly into the base game, Pop'n Music should be updated to support omnimix like IIDX and Museca have (omni flag, importing omnimix data into omni-22 instead of regular, etc).

xinrin:
yeah works like that, but popnhax send a flag and its like this M39:J:A:X:datecode

xinrin:
omni flag added

DragonMinded:
Merged in #73.

PR #86, opened 2024-02-06
SuperNova 2 Support

DiscoStarslayer:
I'm not the best pythoner in the world so some of this code might be cringe, however it is working properly for the following:

Features:
- e-Amuse card registration and login
- Last song selection storage
- Options storage
- Score storage
- Calorie storage
- Events

I still haven't done the following work:

Todo:
- Expose functionality to frontend UI config (calorie display, events)
- ~~Identify why calorie values are not loading~~ Done
- ~~Identify why last song not storing correctly for sn2 category~~ Done
- Doubles mode and course mode logic
- ~~Traffic gen implementation~~ Partially done. Logic for dobules and course mode still TODO so it's not simulated, but singles play is done
- ~~Unit Tests~~ Basic tests added around a quirk regarding user options

Shoutout to @987123879113 for sharing their research with me on the account binary format and the event logic, was a big help in making this work faster.

I plan on getting this a bit more cleaned up and upstream-able eventually, tackling the todos over time, however the branch is fully functional so it would be a shame to keep it to myself till then.

Open to any and all suggestions on the lack of quality on my python code, there was some fishy stuff with the inheritance that I'm not a big fan of but unsure what fits best with the project's vision.


bemani/backend/ddr/base.py
```diff
@@ -127,6 +127,13 @@ def format_profile(self, userid: UserID, profile: Profile) -> Node:
         """
         return Node.void("game")
 
+    def format_profile_part(self, userid: UserID, profile: Profile, part: str) -> Node:
```
DiscoStarslayer, line 130: Profile for SN2 is split into 3 binary chunks, where to break this logic down, it being here feels wrong

bemani/backend/ddr/ddrsn2/ddrsn2.py
```diff
@@ -0,0 +1,566 @@
...
+        hit_chart = self.data.local.music.get_hit_chart(self.game, self.music_version, self.GAME_MAX_SONGS)
+        counts_by_reflink = [0] * self.GAME_MAX_SONGS
+        for reflink, plays in hit_chart:
+            if reflink >= 0 and reflink < self.GAME_MAX_SONGS:
+                counts_by_reflink[reflink] = plays
+        game.add_child(Node.u32_array("cnt_music", counts_by_reflink))
+
+        return game
+
+    def handle_game_hiscore_request(self, request: Node) -> Node:
```
DiscoStarslayer: A lot of this stuff is most likely unused, I used the ddrx code as a baseline, need to take a look

bemani/backend/ddr/ddrsn2/ddrsn2.py
```diff
@@ -0,0 +1,566 @@
...
+        refid = request.attribute("ref_id")
+
+        self.put_profile_by_refid(refid, request)
+
+        root = Node.void("player")
+        root.add_child(Node.s8("result", 2))
+
+        return root
+
+    def handle_info_message_request(self, request: Node) -> Node:
```
DiscoStarslayer: Unsure where this API is used, figure it might get displayed to the user somehow on login but I couldn't work it out

bemani/backend/ddr/ddrsn2/ddrsn2.py
```diff
@@ -0,0 +1,566 @@
...
+        return root
+
+    def handle_player_touch_request(self, request: Node) -> Node:
+        root = Node.void("player")
+
+        root.set_attribute("id", "1234")
+
+        return root
+
+    def handle_info_tenpo_request(self, request: Node) -> Node:
```
DiscoStarslayer: Looks to be some kind of remote-update functionality

bemani/backend/ddr/ddrsn2/playerinfo.py
```diff
@@ -0,0 +1,92 @@
...
+
+        idx = 0
+        for entry in profile.get_int_array("gr_s", 5):
+            player.groove_radar[idx] = entry
+            idx += 1
+
+        idx = 0
+
+        # Empty option set is a non-zero opt array, unsure how to do this check with validated_dict cleanly?
+        if not profile.has_key("opt"):
```
DiscoStarslayer: Had to add new property to validated_dict to support this, maybe a cleaner way?

bemani/backend/ddr/ddrx.py
```diff
@@ -0,0 +1,13 @@
+from typing import Optional
+
+from bemani.backend.ddr.base import DDRBase
+from bemani.backend.ddr.ddrsn2 import DDRSuperNova2
+from bemani.common import VersionConstants
+
+
+class DDRX(DDRBase):
```
DiscoStarslayer, line 8: I had to extract this to fix a circular dependency? Maybe not needed?

bemani/data/mysql/music.py
```diff
@@ -105,8 +114,37 @@ def __get_musicid(self, game: GameConstants, version: int, songid: int, songchar
             },
         )
         if cursor.rowcount != 1:
-            # music doesn't exist
-            raise Exception(f"Song {songid} chart {songchart} doesn't exist for game {game} version {version}")
+            if game == GameConstants.DDR and version == VersionConstants.DDR_SUPERNOVA_2:
```
DiscoStarslayer: Not sure of a clean way to extract the song list from the game yet, so making them on demand here. There is probably a way to do it but the files on the hdd are encrypted so need to think of a method of extracting them.

DragonMinded:
Hey, thanks for sending this in, this looks promising! Sorry about the slow response, I'm very burned out. I won't be able to merge this without at minimum traffic tests, and unit tests would be a nice bonus (but most non-core stuff doesn't have any since unit testing game responses tends to end up being a change inhibitor). Let me comment inline with your comments.

bemani/backend/ddr/base.py
DragonMinded: Honestly the `format_profile` and `unformat_profile` convention were both established to just give some sort of unity across games. They don't need to exist outside of the fact that the base of most games assumes it can call that for profile loads. If there's 3 parts here, this can diverge for cleaner code.

bemani/backend/ddr/base.py
```diff
@@ -142,6 +149,24 @@ def unformat_profile(self, userid: UserID, request: Node, oldprofile: Profile) -
         """
         return oldprofile
 
+    def get_profile_by_refid_and_part(self, refid: Optional[str], part: Optional[str]) -> Optional[Node]:
```
DragonMinded, line 152: Yeah, this definitely feels weird. But, the pattern is established. Worth a think.

bemani/backend/ddr/ddrsn2/__init__.py
```diff
@@ -0,0 +1,3 @@
+from bemani.backend.ddr.ddrsn2.ddrsn2 import DDRSuperNova2
```
DragonMinded: Interesting separating this into a module instead of a single file. Any reason for having this across a bunch of files? Don't see a problem with it, just curious.

bemani/backend/ddr/ddrsn2/ddrsn2.py
DragonMinded: Yeah, I would check whether these are even in the game at all, I would doubt it.

bemani/backend/ddr/ddrsn2/ddrsn2.py
DragonMinded: This could be a string-encoded server message for setting flags? Dunno, that's how bishi does things somewhat.

bemani/backend/ddr/ddrsn2/playerinfo.py
DragonMinded: A `ValidatedDict` is just a subclass of a python dict, so you can just do `if "opt" not in profile` or similar. All of the functions on it are for access or insertion of storage where you want to constrain the type. Under the hood, it's just a `Dict[str, object]`

bemani/backend/ddr/ddrx.py
DragonMinded: IDGI, where's the circular dependency? Does SN2 reference X for some reason? This isn't a problem anywhere else.

bemani/data/mysql/music.py
DragonMinded: You'll need to figure that out, or provide a song DB .tsv or something, because on-demand music stuff isn't supposed to be here. That's why this module has no way of getting the next ID. From it's perspective, the music DB is read-only.

I assume this also means no front-end? I think you mentioned that.

DiscoStarslayer:
> Hey, thanks for sending this in, this looks promising! Sorry about the slow response, I'm very burned out.

Hey! Thank you very much for taking the time to review! I honestly didn't think I would get a response so quickly, so thanks a lot for that.

And understood with the requirements for merge, I agree with that and will work towards polishing this branch up. Game is already 17 years old so there is no urgency to any of this 😄 

Going over your review now 👀

bemani/backend/ddr/ddrsn2/ddrsn2.py
DiscoStarslayer: Yeah the game logic seems to store the returned value to memory, but I haven't dug much further than that, I'll take a peek at the Bishi code, that might give me a few clues.

bemani/backend/ddr/ddrsn2/playerinfo.py
DiscoStarslayer: Thank you! Yes that is much cleaner and much more pythonic, I was looking for a class function but wasn't aware of the `in` syntax. 🙇

bemani/backend/ddr/ddrx.py
DiscoStarslayer: I might try my hand at it again because I think it might have had something to do with how I was importing modules instead of a file directly? But it was mainly a problem with the `previous_version` function.

How the code used to be:
In Stub: `DDRX`, `DDRSN1`
Own Module: `DDRSN2`

`DDRX` has a `previous_version` that points to `DDRSN2` from the module, but `DDRSN2` has a `previous_version` that points to `DDRSN1` from the stub file. So the stub file needs the `DDRSN2` import to work, but that import can't work until the stub file resolves for `DDRSN1`? That's how I understood it but my imports might have been strange too, so let me review this one a bit more.

bemani/data/mysql/music.py
DiscoStarslayer: Yeah currently no front-end, but it's something I want to add so that the events can be managed among other things.

I wasn't aware of the .tsv files, this might be the approach I'll have to take here, there isn't any public tools for decrypting the dnas blobs that I know of and the easiest way is most likely to extract the data from memory once loaded.

bemani/backend/ddr/ddrsn2/__init__.py
DiscoStarslayer: Honestly, mostly just a leftover from my JS work. A common lint is only one class defined per file. If it's a bit awkward I'm ok with merging them into one file.

bemani/backend/ddr/ddrx.py
DragonMinded: Oh, it's because SN1 and X were in the stub file, so circular dependency. You moved the DDRX bits out of the stub into their own file, so it should work now.

bemani/backend/ddr/ddrsn2/__init__.py
DragonMinded: IDGAF either way, pattern in the repo is one file, we don't do silly things like have an import system that depends on there being one export/class per file, so you can get away with a lot of stuff.

AAGACL:
Hello, sry for the bump but I was wondering if we can test this somehow, I havent found any docs or guide how to setup this locally.

bemani/data/mysql/music.py
DiscoStarslayer: @DragonMinded been awhile, but I've cracked the encryption the game uses and have a proper MDB dump from the APP.BIN. Sadly the game doesn't store artist information like the other titles. Is it ok to leave that as None? I see the database allows it but wasn't sure if that would be a showstopper or not. Maybe a tsv could fill that gap if it's needed?

DragonMinded: Leaving nulls for things you can't extract is fine. That's what the game provides. If you want to provide a gap TSV that's what we do for some IIDX stuff that's wrong in the `music_data.bin` files, but it's not necessary.

PR #87, opened 2024-03-29
Hello Pop'n Music support

xinrin:


bemani/frontend/hellopopn/cache.py
```diff
@@ -0,0 +1,19 @@
...
+
+    @classmethod
+    def preload(cls, data: Data, config: Config) -> None:
+        cache = Cache(
+                    app,
+                    config={
+                        "CACHE_TYPE": "filesystem",
+                        "CACHE_DIR": config.cache_dir,
+                    },
+                )
```
DragonMinded: This uses the old preload copy-paste that won't work for anyone using memcached. It should be updated to import the globally available cache like other frontends (see iidx for an example).

bemani/client/hellopopn/hellopopn.py
```diff
@@ -0,0 +1,211 @@
...
+        
+        # Verify pin handling and return card handling
+        self.verify_cardmng_authpass(ref_id, correct=True)
+        self.verify_cardmng_authpass(ref_id, correct=False)
+
+        if ref_id != self.verify_cardmng_inquire(card, msg_type='query', paseli_enabled=paseli_enabled):
+            raise Exception(f'Invalid refid \'{ref_id}\' returned when querying card')
+        self.verify_game_load(ref_id)
+        self.verify_game_load_m(ref_id)
+        self.verify_game_save_m(ref_id)
```
DragonMinded: These don't really test that scores were actually saved, just that the packet worked. See other clients for tests to verify that lower scores don't overwrite higher ones and such.

bemani/backend/hellopopn/hellopopn.py
```diff
@@ -0,0 +1,267 @@
...
+        force_unlock_songs = game_config.get_bool("force_unlock_songs")
+        #if we were on force unlock mode, achievements will not be modified
+        if force_unlock_songs is False:
+          achievements = self.data.local.user.get_achievements(self.game, self.version, userid)
+          love = last.attribute('love')
+          for achievement in achievements:
+                  if achievement.type == 'toki_love' and achievement.id == int(last.attribute('chara')):
+                      love = str(int(achievement.data["love"]) + 1)
+                      if achievement.data["love"] == "5":
+                          love = "5"
```
DragonMinded: I'm confused about the purpose of this. It loads all achievements, overwrites the love given by the client with the achievement that we stored plus one, and then also attempts to cap at 5 in a weird way. What's the intent with this section?

bemani/frontend/hellopopn/endpoints.py
```diff
@@ -0,0 +1,345 @@
...
+    if (
+        re.match(
+            "^["
+            + "\uFF20-\uFF3A"
+            + "\uFF41-\uFF5A"  # widetext A-Z and @
+            + "\uFF10-\uFF19"  # widetext a-z
+            + "\uFF0C\uFF0E\uFF3F"  # widetext 0-9
+            + "\u3041-\u308D\u308F\u3092\u3093"  # widetext ,._
+            + "\u30A1-\u30ED\u30EF\u30F2\u30F3\u30FC"  # hiragana
+            + "]*$",  # katakana
```
DragonMinded: Are the comments supposed to be shifted one down on these?

bemani/utils/read.py
```diff
@@ -4506,6 +4506,154 @@ def import_from_server(self, server: str, token: str) -> None:
...
+        songs = []
+        with open(tsvfile, newline='', encoding="utf-8") as tsvhandle:
+            beatread = csv.reader(tsvhandle, delimiter='|', quotechar='"')  # type: ignore
+            for row in beatread:
+                songid = int(row[0])
+                name = row[1]
+                artist = row[2]
+                easy = 1
+                normal = 10
+                hard = 100
```
DragonMinded: Are these hardcoded constants true for all songs?

bemani/utils/read.py
```diff
@@ -4506,6 +4506,154 @@ def import_from_server(self, server: str, token: str) -> None:
...
+            if song.id not in lut:
+                lut[song.id] = {
+                    'id': song.id,
+                    'title': song.name,
+                    'artist': song.artist,
+                    'genre': song.genre,
+                    'difficulty': {
+                        'easy': 0,
+                        'normal': 1,
+                        'hard': 2,
```
DragonMinded: These difficulties are different than the above ones?

data/hellopopn.tsv
```diff
@@ -0,0 +1,36 @@
...
+26|ふしぎなくすり|上野圭一 feat. SATOE
+27|隅田川夏恋歌|seiya-murai feat.ALT
+28|smooooch・∀・|kors k
+29|DROP THE BOMB|Scotty D.
+30|First Date|ko-saku
+31|jewelry girl*|jun
+32|天庭 おとこのこ編|あさき
+33|murmur twins|yu_tokiwa.djw
+34|恐怖の右脳改革|96
+35|LOVE♥SHINE|小坂りゆ
```
DragonMinded: How was this file derived? Is it like Jubeat where it has to be manually typed? Or was it pulled from the game somehow. Either way, should it not have the difficulties for the songs here?

DragonMinded (commented):
First off, wow, cool! Thank you so much for the new game support, and a weird one at that! I checked this out locally and it introduces a lot of typing issues that looks like it was based off an older revision of code before the cache refactor. It also introduces a fair amount of lint issues. Make sure to run `./verifytyping` and `./verifylint` after rebasing this on the latest changes to see if it is lint/type hint clean. I've also commented on a few issues I saw in the review. Also, I ran the traffic tests to generate some data and some of the pages don't seem to load (scores crashes, records shows no records). This looks pretty close but still needs some work to be accepted.

bemani/utils/read.py
jorpz: hello! has no in game difficulties so this is likely just for sorting purposes? I don't know why there's two different values (as above)

bemani/backend/hellopopn/hellopopn.py
xinrin: since server only sends what chara we level up, server side needs to plus 1, 5 is the max level, so if its more than 5 should stay as 5, ill try to make a cleaner form

bemani/utils/read.py
xinrin: yes there are no difficulties, just for sorting, the two different values is a mistake i made while testing stuff lol, ill fix it

data/hellopopn.tsv
xinrin: i get it from the older revision, and as the other comments, no difficulties in game, i tried to dump the db but looks its based in game files rather in the dll

xinrin:
ill commit fixes tomorrow or day after tomorrow

PR #93, opened 2024-12-27
Update read.py

jorpz:
New glyphs support for POLꓘAMANIA, TYPHØN, and Τέλος (Telos), fix a missed glyph for 和你一起走 (Woneijatheizau, pop'n music song ID 31)


jorpz:
Caught one more, 焱.  There's three more symbols that are unused at this time, two of them for the same potential future song.

PR #94, opened 2025-01-04
Pop'n Music 27 Unilab support

jorpz:
verifylint ✔️
verifytyping ✔️
formatfiles 🍰
trafficgen ✔️
verifytraffic ✔️

No major changes from 26 aside from phase stuff and some new profile stuff.

Known issues:
I don't know how to trigger KAC Lab. This seems to be something that should be able to be accessed on appropriate versions of the dll but I can't seem to figure it out. Rare softlock on pop'n quest Lively II event if you mess with the phase flags and put the game in an invalid state. In theory (and according to bemaniwiki) the entire event should be clearable on earlier Unilab builds.

Not an issue/will not fix:
狼弦暴威 does not appear in Awakening Elem when the event flag is set. The solution to this (for some reason) is to clear the other 10 events. This is not a bemaniutils issue.

### DIFF
--- a/bemani/backend/popn/unilab.py
+++ b/bemani/backend/popn/unilab.py
@@ -535,6 +535,7 @@ class PopnMusicUnilab(PopnMusicModernBase):
     def unformat_profile(self, userid: UserID, request: Node, oldprofile: Profile) -> Profile:
         newprofile = super().unformat_profile(userid, request, oldprofile)
 
+        game_config = self.get_game_config()
         account = request.child("account")
         if account is not None:
             newprofile.replace_int("card_again_count", account.child_value("card_again_count"))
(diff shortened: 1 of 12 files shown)


bemani/backend/popn/unilab.py
```diff
@@ -0,0 +1,640 @@
...
+                team = Node.void("team")
+                event_p27.add_child(team)
+                team.add_child(Node.s16("team_id", team_id))
+                team.add_child(Node.s16("ex_no", ex_no))
+                team.add_child(Node.u32("point", point))
+                team.add_child(Node.bool("is_cleared", is_cleared))
+
+        # Set up Kakusei no Elem achievements
+        game_config = self.get_game_config()
+        if game_config.get_bool("force_unlock_deco"):
```
DragonMinded, line 514: We are setting maximum energy and battery here, but in the `unformat_profile` function we don't ignore updating this value if `force_unlock_deco` is true, which will poison user profiles with fully unlocked events.

bemani/backend/popn/unilab.py
```diff
@@ -0,0 +1,640 @@
...
+                    {
+                        "team_id": team_id,
+                        "ex_no": ex_no,
+                        "point": point,
+                        "is_cleared": is_cleared,
+                    },
+                )
+
+            # Extract Kakusei no Elem achievements
+            battery_data = event_p27.child("battery")
```
DragonMinded, line 622: Should skip this whole section if `force_unlock_deco` is true.

DragonMinded (commented):
Aside from the comments this looks good to merge.

DragonMinded (approved):

PR #101, opened 2025-09-18
Update unilab.py

jorpz:
Adjust max deco id to final value and actually handle the decopa fake "force unlock" that somehow got lost in the shuffle of the older commits


DragonMinded:
It looks like the unformat_profile in common.py would still be broken, as it would persist these on profile save.

DragonMinded:
Note: this is why I don't like having common functionality.

jorpz:
I can split out the whole function. Honestly we should have started doing that with kaimei riddles but I don't want to go back and do that now unless something there is broken stilll

jorpz:
tldr: previous checks to prevent writing to the table seem to work



Before I started messing with anything in common.py I wanted to test to see what the expected behavior vs observed behavior was.

Database was checked for item unlocks and only the deco parts I unlocked naturally were present. Tested a credit with force decopa on, it skips the Awakening Elem event (the check in unilab.py from the initial commits handled this), but at character select all the parts are present as expected from using the force unlock setting.

After the credit finished out and wrote to the db I checked again, db had not changed since the initial check (the force unlocked decopa were not written to the db).

Turned off force unlock, loaded another credit. The previously "unlocked" parts were not present at the character select screen.  Awakening Elem event DID trigger and after the credit an item was unlocked. Just for fun I did another credit and unlocked another part. I took note of the state of the achievements table and those two new items were present.

Then I turned force unlock back on, did another credit. As expected the Awakening Elem screen was skipped, the decopa list was full at chara select. I finished out the credit and the db was still unchanged from the previous check, the two new legit unlocks were there (as were the previous ones) but not overwritten with the force unlocks.

Up to you how you want to move forward with this but it seems like the check in unilab.py that bypasses the deco part unlock event is working to prevent poisoning the achievement database

bemani/backend/popn/unilab.py
```diff
@@ -321,7 +322,16 @@ def get_common_config(self) -> Tuple[Dict[int, int], bool]:
                 # 1 - Net taisen
                 # 2 - Net taisen + Local mode
                 3: 1 if enable_net_taisen else 0,
-                # Unknown event (0-7)
+                # Paseli Festival Attract mode ads (0-7)
+                # As updates to datecodes happen the older ones are are completely locked out
```
jorpz, line 326: It might be cute in the future to check datecode to display the correct values here but eh

bemani/backend/popn/unilab.py
```diff
@@ -391,9 +401,27 @@ def get_common_config(self) -> Tuple[Dict[int, int], bool]:
...
+                #     1992 - Northern Cross
+                # 3 - 1982 - Surf on the Light
+                #     1988 - Danza Pantera
+                #     1983 - バッドエンド・シンドローム
+                # 4 - 1993 - Un Happy Heart
+                #     1987 - 水晶塔のオルカ
+                #     1986 - virkatoの主題によるperson09風超絶技巧変奏曲
+                # 5 - 2017 - virkatoの主題によるperson09風超絶技巧変奏曲 upper
+                # 6 - Event Ended
+                7: popn_quest_lively_2,
```
jorpz, line 420: If you copy/paste these into a different program please make sure you update these to the corrected values.

jorpz:
I've spent too long trying to track down this last phase ID. 
I've looked at delta updates. I've dug around in decompilers. I have no idea. 
I cannot find what, if anything, that last phase ID handles.
I traced and stepforwarded and breakpointed and as far as I can tell phase 13's value is never evaluated.
I played through an entire credit stepping forward in IDA and it just never read.
It was added in the same update as the BPL ad, so my initial guess was it had something to do with that (there are two versions of the screen) but I really have no idea.

I'm sure someone else at some point has dug around in here for other network solutions but as far as I am concerned this ID is just nothing.

Thanks for coming to my TED Talk.

PR #103, opened 2025-10-04
Dance Evolution Arcade Support

DragonMinded:
This adds Dance Evolution Arcade support. Included is support for dance mates, record keeping and profile. The game is shockingly primitive, so attempts cannot be extracted and such records are kept but no individual tries. Note that while DanEvo has no need for rivals or BEMAPI support, support is included anyway. This is mostly useful so that a network can bootstrap support from another network without needing to read local game data.

This PR also serves as a prototypical PR in the case that somebody wants a good example for how to add a new game series to the network code. Some stuff that was done isn't really "up to spec" per-se, such as the scores/records bit, but that's due to DanEvo's shoddy network support and lack of a second version.

### DIFF
--- a/bemani/backend/danevo/danevo.py
+++ b/bemani/backend/danevo/danevo.py
@@ -1,5 +1,6 @@
 import base64
 from typing import Any, Dict
+from typing_extensions import Final
 
 from bemani.backend.ess import EventLogHandler
 from bemani.backend.danevo.base import DanceEvolutionBase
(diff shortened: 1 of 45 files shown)


bemani/backend/danevo/danevo.py
```diff
@@ -0,0 +1,510 @@
...
+    # of attempts.
+
+    # RDATA running score list includes history chunks that are 32 bytes in length. The game
+    # never sends trailing zeros for a data structure, even if it internally recognizes them,
+    # so this may have length not divisible by 32 if there are no non-zero bytes after a certain
+    # location. The correct thing to do is to fill with assumed zero bytes to a 32-byte boundary.
+    # The first 4 bytes of any history chunk are the score in little endian. The next byte is the
+    # song ID. Note that the chart does not appear anywhere in this structure to my knowledge.
+    # I have no idea what the rest of the bytes are, but most stay zeros and many are the same
+    # no matter what.
```
haste, line 39: DanEvo was a fairly fun game to work on tbf. Following should fill in some blanks at least:

```python
    def save_ess_plays(self, profile, rdat01, others, arcade_id):
        if len(rdat01.binary) == 0:
            return

        num_songs = 0
        # Stage 1
        if others["DATA03"].u64s[4] != -1:
            num_songs += 1
        # Stage 2
        if others["DATA03"].u64s[6] != -1:
            num_songs += 1
        # Premium stage
        if others["DATA03"].doubles[0] != -1:
            num_songs += 1

        binary = io.BytesIO(base64.b64decode(rdat01.binary))
        plays = []
        for idx in range(num_songs):
            blob = binary.read(32)
            stage_no = num_songs - idx

            # The oldest entry is missing the last two zero bytes. Other entries just
            # have padding for what's left in the buffer
            (score, params, ts64) = struct.unpack("IIQ", blob[:16].ljust(16, b"\x00"))

            # We've gone too far!
            if score == 0 and params == 0 and ts64 == 0:
                break

            play = DanEvoPlay(
                profile.id,
                timestamp=ts64,
                music_id=params & 0xFF,
                difficulty=params >> 8 & 0xF,
                stage_no=stage_no,
                score=score,
                grade=params >> 27 & 7,
                combo=params >> 12 & 0x3FF,
                fc_type=params >> 30,
                params=params,
            )

            play.arcade_id = arcade_id

            extended = others["DATA04"].strings[stage_no - 1]
            # If we have extended, then fetch data from there. If not, just use
            # the provided params to pull out secret ripples, even if it's really
            # the hiscore one
            if extended != "":
                [
                    _music_id,
                    _music_code,
                    _difficulty,
                    _calories,
                    _grade,
                    _best_grade,
                    secrets,
                    _combo,
                    perfect,
                    great,
                    good,
                    boo,
                ] = extended.split(";")

                play.judge_perfect = int(perfect)
                play.judge_great = int(great)
                play.judge_good = int(good)
                play.judge_boo = int(boo)

                # Values are:
                #  - 0: Not discovered
                #  - 1: Discovered
                #  - 2: Discovered this play
                play.secret_0 = int(secrets[0])
                play.secret_1 = int(secrets[1])
                play.secret_2 = int(secrets[2])
                play.secret_3 = int(secrets[3])
                play.secret_4 = int(secrets[4])
            else:
                play.secret_0 = 1 if params & 0x400000 else 0
                play.secret_1 = 1 if params & 0x800000 else 0
                play.secret_2 = 1 if params & 0x1000000 else 0
                play.secret_3 = 1 if params & 0x2000000 else 0
                play.secret_4 = 1 if params & 0x4000000 else 0
```

DragonMinded: It does, but it also makes me feel like I've completely wasted my time here. There's clearly other people who have worked on this game, but nobody is sharing anything until I put up my own research. Same as it ever was :(

PR #105, opened 2025-11-05
Add legacy decompression to TXP2File

Subject38:
Roughly adapted code from https://github.com/mon/ifs_layeredfs/blob/bc8d504b0db9524f61637ae1fe3455482c02225f/src/texbin.cpp#L1008-L1071. Could probably be cleaned up and also added to the cpp lib for extra speed. I was just trying to mess with mahjong fight club textures but couldn't extract them until adding legacy decompression. I also ran into an issue with the swf parser but I just commented out the lines since I just wanted to extract the txp2 file. Just submitting for upstreaming so if anyone else wants to deal with these files it's one less roadblock :)

### DIFF
--- a/bemani/format/afp/container.py
+++ b/bemani/format/afp/container.py
@@ -372,6 +372,68 @@ class TXP2File(TrackedCoverage, VerboseOutput):
             flags3=flags3,
         )
 
+    def decompress_legacy(self, texture_offset: int, texture_length: int, name: str) -> bytes:
+        # Get size, round up to nearest power of 4
+        decomp_len, comp_len = struct.unpack(
+            ">II",
+            self.data[texture_offset : (texture_offset + 8)],
+        )
+        self.add_coverage(texture_offset, 8)
+        self.vprint(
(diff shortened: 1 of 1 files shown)


DragonMinded:
Huh, yeah, this probably should be an actual LZSS/LZLegacy module with its own compression/decompression and C++ version. It does look very similar to the existing LZ code, maybe the existing could be updated to take a mode flag or something? I haven't looked too hard yet. But also as-is this seems fine because it at least adds one more supported feature. How does this interact with TXP2 repacking? I think that it's smart enough to only rewrite what it needs? But not sure, haven't messed with this in years at this point.

DragonMinded:
Okay, I see the difference. The LZ77 that's used in most modern stuff uses a backref that's relative to the current cursor. That's implemented using a ring buffer, but the important math is that the offset calculated is relative to the current byte. This legacy compression uses an absolute reference into the ring buffer and very carefully initializes the window location to give a lot of zeros to dip into.

A careful modification could make the existing LZ code capable of handling both versions, but I don't have it in me to try and don't have any sample data to work with. Side note, it looks to be identical to the firebeat LZSS decompression that I implemented awhile ago: https://github.com/DragonMinded/firebeat/blob/trunk/utils/lzss.py

PR #106, opened 2025-11-10
[WIP] Beatstream 1 and 2 Support

Hay1tsme:
Funny touchy game. I've been meaning to PR this for literal years but I've never gotten around to it. Currently still need to do frontend, traffic tests, Beatstream 1 events and a few misc things. It's honestly pretty basic as far as network support goes, a lot of "save the value and echo it back" type requests. I should hopefully be able to remove the WIP prefix Soon:tm:.

This also fixes an issue with the example config's webhook section being a list of URLs instead of 1:1 game:url, which seems to be what the code expects. If you'd rather that be a separate PR I can try doing that.

### DIFF
--- a/bemani/backend/bst/__init__.py
+++ b/bemani/backend/bst/__init__.py
@@ -1,0 +1,7 @@
+from bemani.backend.bst.factory import BSTFactory
+from bemani.backend.bst.base import BSTBase
+
+__all__ = [
+    "BSTFactory",
+    "BSTBase",
+]
--- a/bemani/backend/bst/base.py
+++ b/bemani/backend/bst/base.py
@@ -1,0 +1,23 @@
+from bemani.common import Profile
+from bemani.data.types import UserID
+from bemani.backend.base import Base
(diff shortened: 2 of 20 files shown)


jorpz:
Beautiful, this has been on my to-do (with nostalgia) for quite a while and I kept putting it off.  Looking forward to this getting finished!

Subject38:
> Beautiful, this has been on my to-do (with nostalgia) for quite a while and I kept putting it off. Looking forward to this getting finished!

I have nostalgia (force unlock) working but I mainly held off on submitting because I don't want to write traffic tests 🙃. If anyone wants to do the traffic tests and such I'd be down to submit what I have

jorpz:
Oh yes, that would also be fantastic! I could set aside time to probably help with that if it means getting nostalgia 1, forte, and/or op2 (op3 is still n0 so we don't touch it) implemented.  I had put together some of that previously (profiles working, music db loading from the xml) but my SSD died and I never checked it in.

